PR #36, opened 2016-10-24
Issue/23

dchertousov:
Error handling changed. Structure changes.
- Added node-error-helpers dependency
- All errors now have scope of the corresponding services
- All errors are now exported
- Tests updated (and refactored a bit)
- Test function now uses babel
- Eslint airbnb config updated


shmuga:
You forgot to update functions like this https://github.com/Travelport-Ukraine/uapi-json/blob/dev/src/Services/Air/AirService.js#L28
You should look for all dependencies that was previously used.

Issue #29, opened 2016-10-20
Uapi parser should choose version automatically 


shmuga:
Probably it's better to auto-detect version than to define it in code.

https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/uapi-request.js#L100

I think it's better to call uapi parser after xml parser finish it's work. And after that in some way detect correct version of uapi.
https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/uapi-request.js#L108

PR #39, opened 2016-10-25
Uapi version.

shmuga:
- added debug mode to uapi-parser to avoid console.log.
- automatically detecting version of uapi on the step of parsing.
 default version would be 36 and defined in uapi-request.
 after that uapi-parser will make a try to redefine it with response
 version.


shmuga:
Issue #29

PR #58, opened 2016-10-28
M0.5.0

dchertousov:
Milestone description:
- Terminal service added
- Docs updated to reflect Terminal service
- Removed a lot of production and developer dependencies

### DIFF
--- a/src/Services/Terminal/TerminalParser.js
+++ b/src/Services/Terminal/TerminalParser.js
@@ -1,0 +1,48 @@
+import {
+  TerminalRuntimeError,
+  TerminalParsingError,
+} from './TerminalErrors';
+
+const errorHandler = (rsp) => {
+  console.log(rsp);
+  throw new TerminalRuntimeError();
+};
+
+const createSession = (rsp) => {
+  if (
+    !rsp['common_v33_0:HostToken'] ||
+    !rsp['common_v33_0:HostToken']._
+  ) {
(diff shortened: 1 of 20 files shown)


src/Services/Terminal/TerminalParser.js
```diff
@@ -0,0 +1,48 @@
...
+} from './TerminalErrors';
+
+const errorHandler = (rsp) => {
+  console.log(rsp);
+  throw new TerminalRuntimeError();
+};
+
+const createSession = (rsp) => {
+  if (
+    !rsp['common_v33_0:HostToken'] ||
```
shmuga: Since uapi_version is auto-detecting from response I think it's better to use `this.uapi_version` here to support future updates of uapi version

src/Services/Terminal/TerminalParser.js
```diff
@@ -0,0 +1,48 @@
...
+    !rsp['common_v33_0:HostToken']._
+  ) {
+    throw new TerminalParsingError.TerminalSessionTokenMissing();
+  }
+  return rsp['common_v33_0:HostToken']._;
+};
+
+const terminalRequest = (rsp) => {
+  if (
+    !rsp['terminal:TerminalCommandResponse'] ||
```
shmuga: To avoid this check it's better to set root uapi-request function and use uapi-parser instead of default.

shmuga:
Everything LGTM except minor changes.

Issue #62, opened 2016-10-31
Custom log function instead of default.


shmuga:
Add possibility to override default log function to some custom.
Add definition for this function and replace or occurrences of console.log

dchertousov:
Do you want to make it a paramour constructor?
How do you expect the override to work?

shmuga:
There should be function like `function log(message, data)`, that can be passed as parameter in `settings` object. The default value for this function will be `(message, data) => console.log(message,data)`.
So it can be used to push log messages in some specific place (db,file etc.)

shmuga:
Another way we can do it - we can export additional function `setLogFunction()` as external function and set it for whole application.

dchertousov:
And what are you going to use by default if function was not provided? `console.log` or something else?

shmuga:
Yes, exactly.

dchertousov:
I believe constructor option is more convenient way.
Do you want to include it in 0.5.1?

We have plans to release it before wednesday afternoon.

shmuga:
Ok, got it. Will finish till Wednesday.

shmuga:
I think how it would be better to pass log function here.

https://github.com/Travelport-Ukraine/uapi-json/blob/1d65a22a4b2dacccca25cdb68393869aafed2ffe/src/Services/Air/AirParser.js#L326

PR #66, opened 2016-11-03
Added possibility to use custom log function.

shmuga:
Tests for utils.

@dchertousov Could you look on this commit. #62 
I suppose to use it something like this:
```
const { log } = require('./Utils');
log('message', { data: true });
```

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,4 +1,10 @@
+let log = console.log;
+
 const Utils = {
+  setLog: (fn) => {
+    log = fn;
+  },
+  log: (data, message) => log(data, message),
   price(string) {
     if (string === undefined || string === null) {
       return null;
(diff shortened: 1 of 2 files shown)


dchertousov:
I really don't like having global scope log function in utils.
How its going to work with AWS lambda functions?

Maybe, it's a good idea to create terminal services as object instances instead of generated singletones and only then to add this log functionality which then can be easily overrided.

src/utils.js
```diff
@@ -1,4 +1,10 @@
+let log = console.log;
```
dchertousov: As discussed, in this case log function will be global, and creating two services with custom log functions will cause an error.

Me and @shmuga are now discussing another possible ways to restructure services, thus avoiding issue listed above

dchertousov:
@shmuga is it possible to close pull request without merging?

Issue #33, opened 2016-10-20
Plan to addition of FlightService with FlightInformationReq


MiseCZ:
We are planning to add:

```
uAPI.createFlightService(settings) => FlightService
```

With usage like:

```
service.information(params) => Promise
```

Params should be:

```
{
    "airline": "KL",
    "flightNumber": "3098",
    "departure": "2016-10-21"
}
```

Response:

```
{
  "from": "LHR",
  "to": "JFK",
  "departure": "2016-10-21T14:00:00.000+01:00',
  "arrival": "2016-10-21T16:40:00.000-04:00",
  "duration": "PT7H40M",
  "plane": "789",
  "fromTerminal": "3",
  "toTerminal": "4"
}
```

Would that be ok with you or would you suggest otherwise?

shmuga:
Yes! Great idea!
But currently, we are planning to do some refactoring and change some files/code structure.
Look to this [milestone](https://github.com/Travelport-Ukraine/uapi-json/issues/25).
After that, I think you can easily add this Service. 
Also, we are going to provide some Core Api docs and contributing instructions to that time.
Stay tuned!

p.s I think we'll do that till 24 October.

dchertousov:
@MiseCZ What else are you planning to add into FlightService in future?

jirismika:
@dchertousov AFAIK, at this moment we are planning only to add the FlightInformationReq.

I believe we'll deliver the PR next monday (7th October).

MiseCZ:
I believe everything's done here.

dchertousov:
@shmuga , have you planed to fix some documentation?

dchertousov:
@MiseCZ , I can see link to the FlightService at the bottom of README.md
Also @shmuga pointed some tiny issues and planned to fix them

MiseCZ:
Ok. Sorry to be hasty. 

Pls @jirismika how's your plan to correct things?

shmuga:
@MiseCZ I'll take that for me. It's mainly improvements or minor fixes than critical bugs.

shmuga:
Everything fixed in https://github.com/Travelport-Ukraine/uapi-json/commit/ae40b18943643e9162868ad50ec8838b9b1cf7be

PR #70, opened 2016-11-07
Added FlightService

jirismika:
Hello, I've implemented the FlightService as was previously noted in #33 

Options are extended for a required key `key`. Example:
```
options = {
  airline: 'OS',
  flightNumber: '703',
  departure: '2016-11-21',
  key: 'MQ==',
};
```


src/Services/Air/AirErrors.js
```diff
@@ -61,15 +77,22 @@ Object.assign(AirRuntimeError, createErrorsList({
   NoValidFare: 'No valid fare for input criteria.',
   TravelersListError: 'Not all BookingTravelers present in list or wrong lookup keys provided',
 }, AirRuntimeError));
+Object.assign(AirFlightInfoRuntimeError, createErrorsList({
+  FlightNotFound: 'Flight not found',
+  AirlineNotSupported: 'Airline not supported',
+  InvalidFlightNumber: 'Invalid flight number',
+}, AirRuntimeError));
```
dchertousov: Parent of the errors has to be `AirFlightInfoRuntimeError`, not `AirRuntimeError`

shmuga:
Everything looks good to me except new service creation.
It's probably better to put this function to `AirService`.

dchertousov:
@jirismika , do you find it possible to make `flightInfo` function a part of `AirService`?

jirismika:
@dchertousov If you mean to do something like 

```
const airService = uapi.createAirService(settings);
airService.flightInfo(options);
```
... than yes, I do. Is this the final decision to implement it as part of AirService?

Smotrov:
What does this `key` parameter for?

dchertousov:
@jirismika , yes, we see it as a simplification to divide all requests into exactly four types:
- Air
- Hotel
- Utils
- Terminal

jirismika:
Okay, I'm going to move it to AirService.

Key is (to simplify it) a "grouping" parameter :
[The base-64 encoded UUID identifier that links the Flight Information responses to the request criteria. The value that is sent is returned in the resulting FlightInfo(s) elements.](https://support.travelport.com/webhelp/uapi/uAPI.htm#../Subsystems/Schemas/Content/Schemas/FlightInformationReq.html)

dchertousov:
@jirismika , how does the user know which key to send for which request?

Smotrov:
@jirismika sorry, but I can't find any information about "grouping" parameter in provided links.

shmuga:
@jirismika I've looked into WSDL and found that you can pass `<!--1 or more repetitions:-->`  of `FlightInfoCriteria` to this request. Maybe it should be better to pass an array of such options as you mentioned?

Example:
```
[{
  airline: 'OS',
  flightNumber: '703',
  departure: '2016-11-21',
}, {
  airline: 'PS',
  flightNumber: '748',
  departure: '2016-11-21',
} ]
```

And in this case, you can set the `key` property as element index. 
This [line](https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Services/Air/templates/AIR_PRICE_REQ.xml#L25) should help you with that.

And please provide `examples` in the relevant folder for this request. Thanks.

jirismika:
Well, you've actually made me think about it again and I didn't realized earlier, that flightInfo is not returning this key back ... So I also have to extend the response. And as I tested it now, the key can be an empty string, so do you want me to do that parameter optional?

@dchertousov It depends on the user to define the key.

@smotrov 
FlightInformationReq[`key`](https://support.travelport.com/webhelp/uapi/uAPI.htm#../Subsystems/Schemas/Content/Schemas/FlightInformationReq.html)
> The base-64 encoded UUID identifier that links the Flight Information responses to the request criteria. The value that is sent is returned in the resulting FlightInfo(s) elements.

FlightInformationRsp[`CriteriaKey` ](https://support.travelport.com/webhelp/uapi/uAPI.htm#../Subsystems/Schemas/Content/Schemas/FlightInformationRsp.html)
>  	An identifier to link the flightinfo responses to the criteria in request. The value populated here is passed in request.

@shmuga Thanks, you were faster than me. Yes, the `key` is used always in multiple items in response to group/match/link/find the item on request side. Array of requests is not implemented yet, thanks for finding that. Added to my todo list:

TL;DR TODO: 
- question (for all): should be the `key` optional?
- array of requests (and responses)
- examples

Smotrov:
@jirismika I doubt that we should put `key` to interface at all. Even as optional parameter.

jirismika:
Ok, let me ask the question differently: 
Should the `key` be optional or we don't want to return the `key`(s) at all?

Smotrov:
I think we should not return `key` at all.

src/Services/Air/templates/AIR_FLIGHT_INFORMATION_REQUEST.xml
```diff
@@ -6,12 +6,14 @@
     <soapenv:Body>
         <air:FlightInformationReq TargetBranch="{{TargetBranch}}">
             <com:BillingPointOfSaleInfo OriginApplication="uAPI"/>
+            {{#flightInfoCriteria}}
             <air:FlightInfoCriteria
                 Carrier="{{airline}}"
                 DepartureDate="{{departure}}"
                 FlightNumber="{{flightNumber}}"
-                Key="{{key}}"
+                Key=""
```
dchertousov: Does it work with empty key?
Doesn't this make results mixed up?

docs/Air.md
```diff
@@ -177,3 +178,20 @@ If the Universal Record locator code is known, a Universal Record can be retriev
...
+
+**Returns**: <code>Promise</code>
+**See**: [Flight Information](https://support.travelport.com/webhelp/uapi/uAPI.htm#Air/Flight_Information/Flight_Information.htm)
+
+| Param | Type | Description |
+| --- | --- | --- |
+| airline | <code>string</code> | The marketing carrier for the segment. |
+| departure | <code>string</code> | The departure date; either the current date or a date in the future, not include the time zone which is derived from the origin location. |
+| flightNumber | <code>string</code> | The flight number for the segment |
+| key | <code>string</code> | The base-64 encoded UUID identifier that links the Flight Information responses to the request criteria. The value that is sent is returned in the resulting FlightInfo(s) elements. |
```
dchertousov: I think `key` should be removed from the documentation.

src/Services/Air/templates/AIR_FLIGHT_INFORMATION_REQUEST.xml
jirismika: Empty key works fine, there is a new `key` in response, but we are not returning the `key` at all.

dchertousov:
Thank you for your work.

PR #79, opened 2016-11-30
Fixed timeout as optional param

### LINKED ISSUE #78: Terminal examples not working.
I'm trying to use terminal from examples and it's not working.
It's throwing `SessionTimeoutInvalid` error.
But I think that we should use timeout as optional param.

dchertousov:
Fixes #78 

@shmuga , please review

Checking for exact value of `false` instead of `undefined` as if no value is provided timeout is set to `false`.


shmuga:
LGTM
Nice job!

Issue #50, opened 2016-10-27
Support for <common:OverridePCC /> tag


MiseCZ:
We would like to have support for common:OverridePCC tag in AirPriceReq request and possibly in all of AirService requests. I'm not sure about other services.

In our current solution we are  using:
`<com:OverridePCC ProviderCode="{$providerCode}" PseudoCityCode="{$pcc}"/>`

I think:
- The `Settings` object could be probably extended with OverridePCCCode attribute. ProviderCode is probably always 1G. If not OverridePCC should be object.
- The templates from /src/Services/Air/templates should support this setting.

What do you think about it? Thanks.

shmuga:
Looks good for me.

I think you should extend settings object as you mentioned.
We always use provider 1G in our products. So I think we can just add property OverridePCC with value what PCC should we use. After that you should update all xml templates with something like (we are using [handlebars.js](http://handlebarsjs.com/) as template engine)

```xml
{{#if OverridePCC}}
<com:OverridePCC ProviderCode="1G" PseudoCityCode="{{OverridePCC}}"/>
{{/if}}
```

After that you should extend globals in all templates like in this line.

https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Request/uapi-request.js#L64

MiseCZ:
Thanks for reply. I think @jirismika will deliver this function next week.

shmuga:
@dchertousov Has landed last update with `TerminalService` 🎆 .
He added `emulatePcc` property to costructor of this service.
Maybe we can use it in other services as param for pcc overriding?
@Smotrov @MiseCZ  @dchertousov what do you think? Whould it be comfortable creating whole service with overrided pcc or passing it to the function when needed?

dchertousov:
@shmuga looks good for me

jirismika:
Setting `emulatePcc` property using constructor is just fine and I believe we can use it on other services too. 
The question is, which services are supporting it, to know which templates we can start updating. I remember only the UniversalRecordRetrieve and AirPrice supports this, but I believe that it's not the end of the list ...

shmuga:
@jirismika Sorry for so long response. I just missed this one.
As I wrote before you can just put it as an option to the constructor and add to global scope in templates.
We can deprecate old usage for new clear inteface. @dchertousov what do you think?

dchertousov:
@MiseCZ , @jirismika I think we will need this shortly.
[...]

PR #81, opened 2016-12-16
Added support for OverridePCC tag

jirismika:
Hello, 
I've added the OverridePCC tag ( #50 ) to all AirTemplates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ It also has several useful helpers to handle errors.
 
 | Param | Type | Default | Description |
 | --- | --- | --- | --- |
-| auth | `Object <username, password, targetBranch>` | - | See `auth` description [below](#auth). |
+| auth | `Object <username, password, targetBranch, pcc, overridePcc>` | - | See `auth` description [below](#auth). |
 | debug | `Number` | `0` | Can be `0`, `1`, or `2`. |
 | production | `Boolean` | `true` | Production variable is connected with production and pre-production environment. Can be true for production and false for pre-production. For more information read docs. |
 | emulatePcc | `String` | - | Optional. Used for `TerminalService` only. See [`TerminalService`](docs/Terminal.md) |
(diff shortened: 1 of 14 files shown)


shmuga:
Awesome work @jirismika!
I have just requested a review from @dchertousov and after that, we'll merge changes 😎

README.md
```diff
@@ -72,15 +72,16 @@ It also has several useful helpers to handle errors.
 
 | Param | Type | Default | Description |
 | --- | --- | --- | --- |
-| auth | `Object <username, password, targetBranch>` | - | See `auth` description [below](#auth). |
+| auth | `Object <username, password, targetBranch, pcc, overridePcc>` | - | See `auth` description [below](#auth). |
```
dchertousov: I see two new params added in docs: `pcc` and `overridePcc`.
I can see what is `overridePcc` responsible for, but what about `pcc`?
I found it's only used in Air service and only in this line:
https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Services/Air/AirService.js#L25
(and this variable is used then here: https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Services/Air/templates/AIR_CREATE_RESERVATION_REQUEST.xml#L53)

Do you think, guys that `pcc` should really be passed in `auth`?
I can see clear reasons why `overridePcc ` should but not `pcc`.
@shmuga , @jirismika ?

articice: @dchertousov it is the effective pcc of credentials being used. It is copied into each of the search results and not used anywhere else; it is a sort of previous design artefact.
We can remove it, then users should control which pcc they're using by external logic, which fits in perfectly with overridePcc.

Smotrov:
As far as uAPI JSON is made to simplify understanding, my suggestion is to use _emulatePcc_ instead of _overridepcc_. Because is common phrase in industru _to emulate some ones PCC_, but no _to override some ones PCC_.

dchertousov:
@Smotrov 
In my opinion emulating someone's PCC is the same as overriding your own.
In terms of GD SEM command is actually emulating someone's screen, but do we,?

README.md
dchertousov: If it's used only for search results, I think we have to remove it, as any user of library takes control of what's going on.

Smotrov:
Please use only one parameter. Call it **emulatePcc**. Obviously it shouldn't be mandatory.

dchertousov:
@jirismika any updates on `emulatePcc` parameter?

dchertousov:
I'm going to merge this to master as it is with `overridePcc` name, and then change it to `emulatePcc``

PR #92, opened 2017-01-13
Feature/83

### LINKED ISSUE #83: Pass pcc auth param as ticketingPcc book param
I found `pcc` param from `auth` being used only in one transaction in AirService.
I propose to move it from auth to method option.
We should do it along with adding `overridePcc`, because having `pcc` and `overridePcc` in the same context may confuse users.

dchertousov:
Closes #83

Changes:
 - removed `pcc` param from `auth`
 - removed `PCC` from output format for `lowFaresSearch`
 - docs updated
 - examples updated
 - added 3 hours to TAU value (not configurable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ It also has several useful helpers to handle errors.
 
 | Param | Type | Default | Description |
 | --- | --- | --- | --- |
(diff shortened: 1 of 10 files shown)


shmuga:
LGTM

Issue #101, opened 2017-01-17
Bug: incorrect version parsing in errors.


shmuga:
We need to implement "clever" version parsing separately for errors.
Now we have just [return](https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Request/uapi-parser.js#L245) of the same object from parsing. 

```xml
<?xml version="1.0" encoding="utf-8"?>
<SOAP:Envelope xmlns:SOAP="http://schemas.xmlsoap.org/soap/envelope/">
	<SOAP:Body>
		<SOAP:Fault>
			<faultcode>Server.Business</faultcode>
			<faultstring>UNABLE TO FARE QUOTE</faultstring>
			<detail>
				<common_v33_0:ErrorInfo xmlns:common_v33_0="http://www.travelport.com/schema/common_v33_0">
					<common_v33_0:Code>3037</common_v33_0:Code>
					<common_v33_0:Service>AIRSVC</common_v33_0:Service>
					<common_v33_0:Type>Business</common_v33_0:Type>
					<common_v33_0:Description>No availability on chosen flights.</common_v33_0:Description>
					<common_v33_0:TransactionId>AEA29E780A0759C0864E89C1F08A14F9</common_v33_0:TransactionId>
					<common_v33_0:TraceId>admin-5862fc7c7c593</common_v33_0:TraceId>
				</common_v33_0:ErrorInfo>
			</detail>
		</SOAP:Fault>
	</SOAP:Body>
</SOAP:Envelope>
```

PR #103, opened 2017-01-18
Issue/101

shmuga:
Fixing #101 #102

### DIFF
--- a/src/Request/RequestErrors.js
+++ b/src/Request/RequestErrors.js
@@ -19,12 +19,13 @@ Object.assign(RequestValidationError, createErrorsList({
 
 // Runtime errors
 export const RequestRuntimeError = createErrorClass(
-  'RequestRuntimeError',
+  'ParserRuntimeError',
   'Request runtime error',
   errorTypes.RuntimeError
 );
 Object.assign(RequestRuntimeError, createErrorsList({
   TemplateFileMissing: 'XML template not found for request',
+  VersionParsingError: 'Error during parsing version of uapi',
   UnhandledError: 'Error during request. Please try again later',
   ResultsMissing: 'Missing results in response',
 }, RequestRuntimeError));
--- a/src/Request/RequestErrors.js
+++ b/src/Request/RequestErrors.js
@@ -19,7 +19,7 @@ Object.assign(RequestValidationError, createErrorsList({
 
 // Runtime errors
 export const RequestRuntimeError = createErrorClass(
-  'ParserRuntimeError',
+  'RequestRuntimeError',
   'Request runtime error',
   errorTypes.RuntimeError
 );
--- a/src/Request/uapi-parser.js
+++ b/src/Request/uapi-parser.js
@@ -1,11 +1,20 @@
 import _ from 'lodash';
-import Promise from 'promise';
+import xml2js from 'xml2js';
+
 import {
   RequestSoapError,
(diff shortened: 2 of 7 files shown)


src/Request/RequestErrors.js
```diff
@@ -19,12 +19,13 @@ Object.assign(RequestValidationError, createErrorsList({
 
 // Runtime errors
 export const RequestRuntimeError = createErrorClass(
-  'RequestRuntimeError',
+  'ParserRuntimeError',
```
dchertousov: Why do we have this name change here?

dchertousov: As you still keep exporting `RequestRuntimeError`...

src/Request/uapi-parser.js
```diff
@@ -258,6 +284,7 @@ Parser.prototype.parse = function (xml) {
       if (self.debug > 2) {
         console.log('Error during automatic resolving version');
       }
+      throw new RequestRuntimeError.VersionParsingError(obj, e);
```
dchertousov: It's not yet necessary, as we still have fallback version. But I'd like to keep it.
If we decide to keep it, then debug is not needed.

test/parser.test.js
```diff
@@ -1,15 +1,26 @@
 const assert = require('assert');
+const fs = require('fs');
+
 const errors = require('../src').errors;
 const UAPIParser = require('../src/Request/uapi-parser');
 
-const parser = new UAPIParser('v_36_0', {});
+const xmlFolder = __dirname + '/FakeResponses';
```
dchertousov: Please, use `path.join()` instead.

src/Request/uapi-parser.js
shmuga: Let's keep this error throw as a part of fallback version deprecation.

shmuga:
@dchertousov recheck and approve pls

test/parser.test.js
```diff
@@ -1,10 +1,11 @@
 const assert = require('assert');
 const fs = require('fs');
+const path = require('path');
 
 const errors = require('../src').errors;
 const UAPIParser = require('../src/Request/uapi-parser');
 
-const xmlFolder = __dirname + '/FakeResponses';
+const xmlFolder = path.join(__dirname, '/FakeResponses');
```
dchertousov: If you use join, then '/' is not needed.

test/parser.test.js
```diff
@@ -18,7 +19,7 @@ describe('parser tests', () => {
 
   it('should detect error version automaticly', () => {
     const parser = new UAPIParser('someroot', 'v36_0');
-    const xml = fs.readFileSync(`${xmlFolder}/Other/UnableToFareQuoteError.xml`);
+    const xml = fs.readFileSync(path.join(xmlFolder, '/Other/UnableToFareQuoteError.xml'));
```
dchertousov: Same goes here.

src/Request/uapi-parser.js
dchertousov: I'd like to keep it too!
I meant removing these lines:
```javascript
       if (self.debug > 2) {
         console.log('Error during automatic resolving version');
       }
```

As we don't need logging if we are throwing an error

dchertousov:
Waiting for changes discussed in IM

shmuga:
@dchertousov done!

PR #109, opened 2017-01-23
Feature/86

### LINKED ISSUE #86: Add getTicket method
getTicket method should be added to AirService in order to provide detailed information about the ticket, as importuner method gives only ticket number.

dchertousov:
Closes #86 

Added getTicket method


src/Services/Air/AirValidator.js
```diff
@@ -208,6 +208,26 @@ Validator.prototype.flightInfoItem = function (item) {
   }
 };
 
+Validator.prototype.fop = function () {
+  if (!this.params.fop) {
+    throw new AirValidationError.FopMissing();
+  }
+  if ((typeof this.params.fop) !== 'object' || this.params.fop.type !== 'Cash') {
```
shmuga: Please use `Object.prototype.toString.apply({})` instead of typeof

shmuga:
Please provide tests for parser and validator.

PR #112, opened 2017-01-30
getTickets method added

### LINKED ISSUE #94: Add getTickets method for AirService
It would be nice to have a method which will accept PNR number and will return not only the list of tickets but an array of getTicket results for each ticket in PNR

dchertousov:
Closes #94


dchertousov:
The only thing is may be there is a good idea to add .catch for some general errors in method.

dchertousov:
@shmuga , please review

test/Air/AirParser.test.js
```diff
@@ -47,8 +200,10 @@ describe('#AirParser', function () {
       const uParser = new ParserUapi('air:LowFareSearchRsp', 'v33_0', {});
       const parseFunction = airParser.AIR_LOW_FARE_SEARCH_REQUEST;
       const xml = fs.readFileSync(`${xmlFolder}/LowFaresSearch.2ADT1CNNIEVBKK.xml`).toString();
-      const jsonResult = require('../FakeResponses/Air/LowFaresSearch.2ADT1CNNIEVBKK.json');
-      return uParser.parse(xml).then(json => {
+      const jsonResult = JSON.parse(
```
shmuga: What was the reason of replacing require with fs?

dchertousov: Linter companied about requires inside file, not on the top.

shmuga: We are linting only src code. So it's not necessary to refactor tests.

PR #121, opened 2017-02-02
Feature/119

### LINKED ISSUE #119: Add getPNRByTicketNumber method
This method will be a part of `AirService` and will use `TerminalService` in order to obtain PNR from terminal response.

PNR number is required to import PNR before calling getTicket if PNR was not imported and it's impossible to get extended ticketing information for ticket (See #114).

dchertousov:
Added getPNRByTicketNumber method.
Closes #119


src/Services/Air/Air.js
```diff
@@ -81,6 +82,32 @@ module.exports = (settings) => {
...
+      const terminal = createTerminalService(settings);
+      const memo = {};
+      return terminal.executeCommand(`*TE/${options.ticketNumber}`)
+        .then(
+          (response) => {
+            memo.response = response;
+          }
+        )
+        .then(
+          () => terminal.closeSession()
```
shmuga: Can be written shortly as `.then(terminal.closeSession)`.

src/Services/Air/Air.js
```diff
@@ -81,6 +82,32 @@ module.exports = (settings) => {
...
+        )
+        .then(
+          () => {
+            try {
+              return memo.response.match(/RLOC [^\s]{2} ([^\s]{6})/)[1];
+            } catch (err) {
+              throw new AirRuntimeError.PnrParseError(memo.response);
```
shmuga: Is there any reason for separate error type?

dchertousov: In general it is for not throwing

TypeError: Cannot read property '1' of null

src/Services/Air/Air.js
dchertousov: Shortened

src/Services/Air/Air.js
shmuga: Ok. Let it be 👌

src/Services/Air/Air.js
dchertousov: @see #1e19df5

Issue #87, opened 2017-01-12
Universal search for bookings & tickets


dchertousov:
We need to create universal `.search` method for `AirService` which will accept any string, and then will be able to return search results after performing following actions:
 - searching for PNR (if the string contains 6 alphanumeric characters)
 - searching for tickets, when string contains numbers, spaces and dashes only

dchertousov:
We should think about the format of search results as this kind of request may return results of different types.

Smotrov:
Lets call it `.searchBookings`.
To do not mix it up with shopping functions

Smotrov:
Format should be simple (as we love to).
There should be only one input string parameter `searchPhrase`. 
Result should be an object that contains array of PNR's and array of warnings. 

Function should also have not mandatory emulation functional.

dchertousov:
As we also search for tickets, I don't think that array of PNRs is enough :)

shmuga:
I don't think that we even need such function if we have import by PNR and if we are going to implement this function https://github.com/Travelport-Ukraine/uapi-json/issues/86. 
Maybe it's better to have utils module that will export next functions:
- isTicket(ticket: string) -> boolean
- isPNR(pnr: string) -> boolean

So we'll receive simplicity over versatility.

Smotrov:
If user will pass ticket number in `searchPhrase` function should return a PNR that contains this ticket anyway.

dchertousov:
Well, it's not always possible.
At least in case, when booking is already archived, but ticket still exists.

Smotrov:
Oh! Exactly! Then the second array should contain _ticketsWithArchvedBookings_

dchertousov:
I propose just this:

```
{
  bookings: [
    // PNR list
  ],
  tickets: [
    // Tickets list
  ]
}
```

Each ticket will have corresponding pnr attribute

Smotrov:
Please change `tickets` to `ticketsWithArchvedBookings`
Because if someone has passed ticket number as a parameter and this ticket has an active, retrievable booking in this cace the first array should contain booking number of this booking & second should be empty (or not exist at all).

dchertousov:
I find it a little strange when someone enters ticket number, but receives a whole booking. In RDP, for instance, searching by ticket number opens a ticket and gives ability to void or refund it, as ticket is a separate object.

dchertousov:
@Smotrov should it remain in the current milestone, or should we move it to 0.8.0?
[...]

PR #143, opened 2017-02-23
Issue/87

shmuga:
#87


docs/Air.md
```diff
@@ -239,3 +239,32 @@ Request for the ticket information for all tickets in PNR.
...
 
 **See: <a href="../examples/Air/getTickets.js">getTickets example</a>**
+
+
+<a name="searchBookingsByPassengerName"></a>
+### .searchBookingsByPassengerName(params)
+> Requires Terminal access enabled in uAPI. See [TerminalService](Terminal.md)
+Request for the ticket information.
+
+**Returns**: `Promise` with next shape `{ type: 'list' | 'pnr', data: Array | String }`
```
dchertousov: Why do we need such a complication?
Why don't we always return an array? (Of one element if the only booking is found)

src/Services/Air/Air.js
```diff
@@ -29,16 +31,9 @@ module.exports = (settings) => {
...
               LocatorCode: code,
-            }).then(() => {
-              throw err;
-            }).catch(() => {
-              if (settings.debug > 0) {
-                console.log('Cant cancel booking with UR', code);
-              }
-              throw err;
-            });
+            }).then(() => Promise.reject(err));
```
dchertousov: Nicely done

src/Services/Air/Air.js
```diff
@@ -68,21 +63,14 @@ module.exports = (settings) => {
       const parameters = {
         flightInfoCriteria: _.isArray(options) ? options : [options],
       };
-      return service.flightInfo(parameters)
-        .then(data => data)
```
dchertousov: This line is unnecessary

src/Services/Air/Air.js
```diff
@@ -129,5 +117,55 @@ module.exports = (settings) => {
           err => Promise.reject(new AirRuntimeError.UnableToRetrieveTickets(options, err))
         );
     },
+
+    searchBookingsByPassengerName(options) {
+      const terminal = createTerminalService(settings);
+      return terminal.executeCommand(`*-${options.searchPhrase}`)
+        .then((firstScreen) => {
+          const list = parsers.searchPaxList(firstScreen);
```
dchertousov: Could you please change it to `parsers.searchPassengersList`?
We never user PAX before, so, let it be passengers.

src/Services/Air/Air.js
```diff
@@ -129,5 +117,55 @@ module.exports = (settings) => {
...
+                          firstScreenAgain,
+                        })
+                      );
+                    }
+
+                    return localTerminal.executeCommand(`*${line.id}`);
+                  })
+                  .then(parsers.bookingPnr)
+                  .then(pnr => localTerminal.closeSession()
+                    .then(() => ({ ...line, pnr }))
```
dchertousov: Could be possibly changed to:
```javascript
.then(pnr => localTerminal.closeSession())
.then(() => ({ ...line, pnr }));
```

just to not wrap around once more

src/Services/Air/Air.js
```diff
@@ -129,5 +117,55 @@ module.exports = (settings) => {
...
+                  );
+              }))
+              .then(data => ({ type: 'list', data }));
+          }
+
+          const pnr = parsers.bookingPnr(firstScreen);
+          if (pnr) {
```
dchertousov: Once again, I'm voting to return always array.
It will simplify using this call for the 3rd party like this:
```javascript
AirService.searchBookingsByPassengerName({searchPhrase: 'IVANOV'})
  .then(results => results.forEach((item) => { /* do something */ });
```
instead of checking type of response

docs/Air.md
shmuga: That's because we don't parse booking screen completely and don't have any information about flight date and passenger names.

src/Services/Air/Air.js
shmuga: Ok

src/Services/Air/Air.js
shmuga: Nope than you'll lose context and pnr variable.

src/Services/Air/Air.js
shmuga: It's already deleted 😸

src/Services/Air/Air.js
dchertousov: Oh, I see

Issue #140, opened 2017-02-14
Add cancelTicket method


dchertousov:
As we are working in #84 on cancelling bookings and cancelling tickets is a part of it, I propose to have separate `cancelTicket` method. It will utilize the same `AirVoidDocumentReq`.

Issue #84, opened 2017-01-06
Canceling booking


Smotrov:
There is a need for method that will do a canceling of the existing bookings.
As usually it would be perfect to simplify an interface as it possible. I guess that one parameter _PNR number_ should be more then enough.

Smotrov:
Canceling should be done [by PNR](https://support.travelport.com/webhelp/uapi/uAPI.htm#Booking/Cancel/Canceling_Bookings.htm).

dchertousov:
@Smotrov I think we should check for open tickets before cancelling and if any return an `Error`

dchertousov:
I propose following options for method:

```json
{
  pnr: 'PNR001',
  force: true
}
```

If `force` param passed and equals `true`, then uapi-json tries to cancel tickets.
If no `force` params or it doesn't equal `true`, then uapi-json throws an error if there are tickets with `OPEN` status.

dchertousov:
@Smotrov , please confirm

Smotrov:
Please use `cancelAllTicketsInPNR` instead of `force`.

dchertousov:
Too long, may it be just, `cancelTickets` ?

Smotrov:
Ок

PR #153, opened 2017-02-28
Feature/84

dchertousov:
Added cancelTicket and cancelPNR methods

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -169,19 +169,7 @@ module.exports = (settings) => {
     },
 
     cancelTicket(options) {
-      return service.getTicket(options)
-        .catch(
-          (err) => {
-            if (!(err instanceof AirRuntimeError.TicketInfoIncomplete)) {
-              throw err;
-            }
-            return this.getPNRByTicketNumber({
-              ticketNumber: options.ticketNumber,
-            })
-              .then(pnr => this.importPNR({ pnr }))
-              .then(() => service.getTicket(options));
-          }
-        )
+      return this.getTicket(options)
         .then(
           ticketData => service.cancelTicket({
             pnr: ticketData.pnr,
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -225,14 +225,15 @@ module.exports = (settings) => {
                     ticket.coupons[0].status !== 'V' ? service.cancelTicket({
                       pnr: options.pnr,
                       ticketNumber: ticket.ticketNumber,
-                    }) : true
+                    }) : Promise.resolve(true)
                   )
                 )
               );
             }
           ))
         )
-        .then(() => service.cancelPNR(options))
+        .then(() => this.importPNR(options))
+        .then(pnrData => service.cancelPNR(pnrData[0]))
         .catch(
           err => Promise.reject(new AirRuntimeError.FailedToCancelPnr(options, err))
         );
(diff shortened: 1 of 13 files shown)


src/Services/Air/Air.js
```diff
@@ -43,14 +43,14 @@ module.exports = (settings) => {
     },
 
     ticket(options) {
-      return service.importPNR(options).then((data) => {
+      return this.importPNR(options).then((data) => {
```
shmuga: Be careful because using `this` can cause troubles in test. Don't forget to mock service functions after that.

src/Services/Air/Air.js
```diff
@@ -167,5 +167,77 @@ module.exports = (settings) => {
             .catch(() => results)
         );
     },
+
+    cancelTicket(options) {
+      return service.getTicket(options)
```
shmuga: This part looks like duplication of functionality provided in `Air.getTicket`

src/Services/Air/Air.js
dchertousov: Air test actually rely on mocking service functions.

src/Services/Air/Air.js
```diff
@@ -167,5 +167,77 @@ module.exports = (settings) => {
...
+                  coupon => 'OV'.indexOf(coupon.status) === -1
+                )
+              );
+              if (hasNotOpenSegment) {
+                return Promise.reject(new AirRuntimeError.UnableToCancelTicketStatusNotOpen());
+              }
+              return Promise.all(
+                ticketData.tickets.map(
+                  ticket => (
+                    ticket.coupons[0].status !== 'V' ? service.cancelTicket({
```
shmuga: Maybe better to move this checks to cancelTicket request and avoid code duplication here.

src/Services/Air/AirParser.js
```diff
@@ -326,12 +319,22 @@ const airGetTicket = function (obj) {
       };
     }
   );
-  const taxes = Object.keys(airPricingInfo['air:TaxInfo']).map(
-    taxKey => ({
-      type: airPricingInfo['air:TaxInfo'][taxKey].Category,
-      value: airPricingInfo['air:TaxInfo'][taxKey].Amount,
-    })
-  );
+  const taxes = airPricingInfo !== null ? (
```
shmuga: I think it's more compact to write `airPricingInfo && ... `

src/Services/Air/AirParser.js
```diff
@@ -439,7 +457,7 @@ function extractBookings(obj) {
       }
     );
 
-    const reservations = Object.keys(booking['air:AirPricingInfo']).map(
+    const reservations = !booking['air:AirPricingInfo'] ? [] : Object.keys(booking['air:AirPricingInfo']).map(
```
shmuga: Please write ternary operators multi-line. It's more readable.

src/Services/Air/Air.js
dchertousov: Agree

src/Services/Air/Air.js
dchertousov: No, this checks are to detect are all of the tickets eligible to be canceled. If we move this to cancelTicket we may run into the situation where some tickets are cancelled, and some are not.

dchertousov:
@shmuga updated

dchertousov:
Closes #140, closes #84

dchertousov:
#140

PR #162, opened 2017-03-07
Exchange quote

shmuga:


src/Services/Air/AirErrors.js
```diff
@@ -106,6 +107,8 @@ Object.assign(AirRuntimeError, createErrorsList({
   UnableToOpenPNRInTerminal: 'Unable to open requested PNR in Terminal',
   UnableToAddExtraSegment: 'Unable to add extra segment to PNR',
   UnableToSaveBookingWithExtraSegment: 'Unable to save booking with extra segment',
+  NoResidualValue: 'Original ticket has no residual value for this specific itinerary. Issue a new ticket using current fares.',
+  TicketsNotIssued: 'Host error during ticket retrieve.',
```
dchertousov: Is it retrieve or issue?

src/Services/Air/AirParser.js
```diff
@@ -576,6 +580,103 @@ function gdsQueue(req) {
...
+  };
+
+  return utils.deflate(JSON.stringify(token))
+    .then((zippedToken) => {
+      const json = this.mergeLeafRecursive(req, root)['air:AirExchangeQuoteRsp'];
+
+      const exchangeInfoRoot = `common_${this.uapi_version}:AirExchangeInfo`;
+
+      const exchangeBundleTotal = json['air:AirExchangeBundleTotal'];
+      const totalBundle = exchangeBundleTotal
```
dchertousov: Is this a boolean value? You use in line 673 as not boolean

src/Services/Air/AirParser.js
```diff
@@ -576,6 +580,103 @@ function gdsQueue(req) {
...
+              const fare = pricing['air:FareInfo'][info.FareInfoRef];
+
+              return {
+                bookingCode: info.BookingCode,
+                cabinClass: info.CabinClass,
+                baggage: format.getBaggage(fare['air:BaggageAllowance']),
+                fareBasis: fare.FareBasis,
+                from: fare.Origin,
+                to: fare.Destination,
+                uapi_segment_ref: info.SegmentRef,
```
dchertousov: Should it be `uapi_booking_ref` or something instead?

src/Services/Air/AirParser.js
```diff
@@ -576,6 +580,103 @@ function gdsQueue(req) {
...
+              };
+            });
+
+          return {
+            ...format.formatPrices(pricing),
+            bookingInfo,
+            uapi_pricing_info_ref: pricing.Key,
+            fareCalc: pricing['air:FareCalc'],
```
dchertousov: We already use `fareCalculation` in few places for this value.

src/Services/Air/AirParser.js
```diff
@@ -576,6 +580,103 @@ function gdsQueue(req) {
...
+          pricingType: airPricingDetails.PricingType,
+          lowFareFound: airPricingDetails.LowFareFound === 'true',
+          lowFarePricing: airPricingDetails.LowFarePricing === 'true',
+          discountApplies: airPricingDetails.DiscountApplies === 'true',
+          penaltyApplies: airPricingDetails.DiscountApplies === 'true',
+          conversionRate: parseFloat(airPricingDetails.ConversionRate || 1),
+          rateOfExchange: parseFloat(airPricingDetails.RateOfExchange || 1),
+          validatingVendor: airPricingDetails.ValidatingVendorCode,
+        },
+        pricingSolution: { ...format.formatPrices(solution) },
```
dchertousov: Why
```javascript
pricingSolution: { ...format.formatPrices(solution) },
```

instead of just

```javascript
pricingSolution: format.formatPrices(solution),

File: src/Services/Air/AirValidator.js [hunk 6]
```diff
@@ -14,7 +14,9 @@ module.exports = {
   ),
 
   AIR_PRICE: compose(
-    validate(),
+    validate(
+      validators.segments,
```
dchertousov: Great job about segments validation!

src/utils/index.js
```diff
@@ -7,6 +7,8 @@ import parsers from './parsers';
 import transform from './transform';
 import compose from './compose';
 import hasAllFields from './has-all-required-fields';
+import inflate from './inflate-promise';
```
dchertousov: Nice to have this features

src/Services/Air/AirErrors.js
shmuga: This error occurs when tickets are not issued and this is an original system response.

src/Services/Air/AirParser.js
shmuga: Right! I'll change this one.

src/Services/Air/AirErrors.js
dchertousov: Ok, let's leave it for now

src/Services/Air/AirParser.js
shmuga: && operator returns value of last evaluated argument
```
var p = null
p && p.params // will not throw error and return null

var p = { params: 123 }
p && p.params // will be reslved as 123
```

shmuga: So it's like shorthand for 
(null) ? 123 : null

src/Services/Air/AirParser.js
shmuga: But this is a reference to segment. How should we name it?

src/Services/Air/AirParser.js
shmuga: That's the equivalent syntax. I think we can use both.

src/Services/Air/AirParser.js
dchertousov: Well, I missed, that you then use the property of the same object.

src/Services/Air/AirParser.js
dchertousov: So the whole booking info has a reference to a single segment?

src/Services/Air/AirParser.js
dchertousov: It is an additional call for Object.assign
I'd like to have it removed until we will have more to add into this object

Issue #127, opened 2017-02-07
Implement Air Exchange function from UAPI.


shmuga:

shmuga:
Partly released in #189 
Need to be reopened in future for improvements for many edge cases.

PR #189, opened 2017-03-22
Issue/127

shmuga:


src/Services/Air/AirErrors.js
```diff
@@ -109,6 +111,8 @@ Object.assign(AirRuntimeError, createErrorsList({
   UnableToSaveBookingWithExtraSegment: 'Unable to save booking with extra segment',
   NoResidualValue: 'Original ticket has no residual value for this specific itinerary. Issue a new ticket using current fares.',
   TicketsNotIssued: 'Host error during ticket retrieve.',
+  CantDetectExchangeReponse: 'Exchange response is unknown.',
+  ExchangeTokenIncorrect: 'Cant parse exchange token. Please resolve it again.',
```
dchertousov: Please, change `Cant` to `Can not` or to `Can\'t`

src/Services/Air/transformers/index.js
```diff
@@ -4,6 +4,7 @@ import setPassengersAge from './set-passengers-age';
 import setHasFareBasisFlag from './set-has-farebasis-flag';
 import setGroupsForSegments from './set-groups-for-segments';
 import addMetaPassengersBooking from './add-meta-passengers-booking';
+import decodeExchangeToken from './decode-exchange-token-and-build';
```
dchertousov: What does `-and-build` mean?
Use it or remove it.

src/Services/Air/validators/reservation-locator.js
```diff
@@ -0,0 +1,8 @@
+import { AirValidationError } from '../AirErrors';
+
+export default (params) => {
+  if (params.uapi_reservation_locator === undefined
```
dchertousov: I'm not sure about format. Is it base64-encoded? You may check it with regexp then. If not, that's ok.

src/Services/Air/transformers/index.js
shmuga: It means to build xml as result. Should I change it to `build-xml`?

src/Services/Air/validators/reservation-locator.js
shmuga: I'm not sure it's exactly base64, maybe we just should check if it's string and has length more than one?

src/Services/Air/transformers/index.js
dchertousov: I mean you use it everywhere as just `decodeExchangeToken`.
I think you could name the file just `decode-exchange-token` then.

src/Services/Air/validators/reservation-locator.js
dchertousov: Aha, that's not token, it's a string from uapi.
Yes, just check for a string here.

src/Services/Air/validators/exchange-token.js
```diff
@@ -0,0 +1,8 @@
+import { AirValidationError } from '../AirErrors';
+
+export default (params) => {
+  if (params.exchangeToken === undefined
```
dchertousov: Here you should check either for string or base64-string, if you'd like to

shmuga:
Fixes #127

PR #194, opened 2017-04-06
Issue/193

dchertousov:
Temporary workaround for issue with missing bookingInfo fragments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "0.8.7",
+  "version": "0.8.8",
   "description": "Travelport Universal API",
   "main": "build/",
   "files": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "0.8.6",
+  "version": "0.8.8",
```
shmuga: why double version up? is stable not merged to master?

dchertousov: https://www.npmjs.com/package/uapi-json
We had 0.8.7 somehow.
Stable not yet merged, because we are making branches from stable if we fix something.
And only then stable will be merged to master

dchertousov: Changed base to stable

PR #200, opened 2017-05-20
Added terminal commands logging

dchertousov:
Added terminal commands logging

### DIFF
--- a/src/Services/Terminal/Terminal.js
+++ b/src/Services/Terminal/Terminal.js
@@ -17,6 +17,7 @@ module.exports = function (settings) {
   const service = terminalService(settings);
   const emulatePcc = settings.auth.emulatePcc || false;
   const timeout = settings.timeout || false;
+  const debug = settings.debug || 0;
   const state = {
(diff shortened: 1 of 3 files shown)


src/Services/Terminal/Terminal.js
```diff
@@ -95,6 +103,9 @@ module.exports = function (settings) {
       .then(processResponse)
       .then(
         (response) => {
+          if (debug) {
```
shmuga: By the way, we have different logging modes. 
Let's use them as described in docs in README.md
So here you need to check for `debug === 2`. And check all console.log statements in this PR.

dchertousov: In serverless setup for TCP-air we have debug more set to 1. I wanted to keep this output for debug level 1, as in debug level 2 we have all the XML output. Let's keep it like that. I'd like to have commands output for any debug level

shmuga: Deal 🤝

PR #205, opened 2017-05-25
Issue/204

dchertousov:
Changes for NO ADC tickets.  …
- TotalPrice = 0
- noAdc = true

### DIFF
--- a/src/Services/Air/AirParser.js
+++ b/src/Services/Air/AirParser.js
@@ -360,13 +360,18 @@ const airGetTicket = function (obj) {
         })
       )
     : [];
-  const priceInfo = {
-    totalPrice: etr.TotalPrice,
+  const priceInfo = Object.assign({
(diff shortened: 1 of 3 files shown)


src/Services/Air/AirParser.js
```diff
@@ -360,13 +360,18 @@ const airGetTicket = function (obj) {
...
-  const priceInfo = {
-    totalPrice: etr.TotalPrice,
+  const priceInfo = Object.assign({
+    totalPrice: etr.TotalPrice || 0,
     basePrice: etr.BasePrice,
     equivalentBasePrice: etr.EquivalentBasePrice,
     taxes: etr.Taxes,
     taxesInfo: taxes,
-  };
+  }, etr.TotalPrice
```
shmuga: Maybe just set it to false? In such way we will support one shape of object?

PR #210, opened 2017-06-20
Added parsing of commission and tests.

shmuga:


dchertousov:
`uapi` does not return any commission notices when commission amount is selected for ticket.
Therefore we could not return this information.

Issue #242, opened 2017-07-26
Add FOP type Card for .ticket()


articice:
Colleagues, I've added the card payment for .ticket() method.

The request format is this:
```
  fop: {
    type: 'Card',
  },
  creditCard: {
    number: '5437****',
    name: 'NAME ON CARD',
    expDate: '01/25',
    cvv2: '123',
    //type: 'CA',
  },
```

I've also added [Joi](https://www.npmjs.com/package/joi) JSON schema validation, instead of writing it as native code and using a separate Luhn algorithm check.

I suggest transforming all validators to use this neat module, it creates verbose data validation messages.

There are some notes:
- the optional `type` value is not passed into CreditCard object; it can be specified optionally by the user as one of 1G card types - `CA`, `AX`, `JC`, `VI`, `DC`.
- card data is passed as `CreditCard` object, not `DebitCard`. They are identical in structure. I'm not sure if any auto-detection is possible and why the API has these two different container types.
- I was unable to issue tickets in pre-prod environment because of a plating carrier error, will do further testing.

dchertousov:
@articice , please create PR

dchertousov:
@articice , please fix PR, according to the hints, provided by tests and lint processors

PR #243, opened 2017-07-26
#242 add FOP type Card for .ticket()

articice:
- use Joi validation
- pass as CreditCard object (not DebitCard)
- type param (two letter 1G card type) is optional


shmuga:
@articice I think you also should cover transformer you've added.

PR #244, opened 2017-07-26
Moved some information to upper level.

shmuga:


src/Services/Air/AirParser.js
```diff
@@ -540,33 +540,41 @@ function extractBookings(obj) {
...
 
+    const fareQuotesCommon = {};
+
     const pricingInfos = !booking['air:AirPricingInfo']
       ? []
       : Object.keys(booking['air:AirPricingInfo']).map(
         (key) => {
-          const reservation = booking['air:AirPricingInfo'][key];
-          const uapiSegmentRefs = reservation['air:BookingInfo'].map(
+          const fareQuote = booking['air:AirPricingInfo'][key];
```
dchertousov: Actually, it's not fareQuote, it's pricingInfo.

src/Services/Air/AirParser.js
```diff
@@ -540,33 +540,41 @@ function extractBookings(obj) {
       })
       : [];
 
+    const fareQuotesCommon = {};
```
dchertousov: It's common data not for fare quotes, but for pricingInfos, which will be distributed between fareQuotes

PR #246, opened 2017-07-29
Added custom logging to all services.

shmuga:


src/Request/uapi-request.js
```diff
@@ -41,7 +47,7 @@ module.exports = function (service, auth, reqType, rootObject,
 
   return function serviceFunc(params) {
     if (debugMode) {
-      console.log('Input params ', params);
+      log('Input params ', pd.json(JSON.stringify(params)));
```
dchertousov: It's better to pass JSON untouched just as `pd.json(params)`: https://github.com/vkiryukhin/pretty-data/blob/master/pretty-data.js#L137
Otherwise it will be parsed and stringified again.

src/Request/uapi-request.js
```diff
@@ -85,15 +91,22 @@ module.exports = function (service, auth, reqType, rootObject,
       })
         .then((response) => {
           if (debugMode > 1) {
-            console.log('Response SOAP: ', response.data);
+            log('Response SOAP: ', pd.xml(response.data));
```
dchertousov: Is there always a valid XML in non-error response?

shmuga: It would not be parsed if it's not and error would be thrown.

src/Request/uapi-request.js
shmuga: Cool! Will do that.

src/Request/uapi-request.js
dchertousov: Ok! So we could leave it this way.

src/Request/uapi-request.js
dchertousov: 👍

shmuga:
Closes #62

PR #256, opened 2017-08-04
Added check for string

shmuga:


dchertousov:
Please, update version before merge and release the fix.

PR #260, opened 2017-08-08
Fixed error with order of fare quotes.

shmuga:


dchertousov:
Please, increase version before approval.

PR #264, opened 2017-08-14
Adde parsing of split bookings

dchertousov:


src/Services/Air/AirParser.js
```diff
@@ -709,29 +742,33 @@ function extractBookings(obj) {
...
+        modifiedAt: providerInfo.ModifiedDate,
+        fareQuotes,
+        segments,
+        serviceSegments,
+        passengers,
+        emails,
+        bookingPCC: providerInfo.OwningPCC,
+        tickets,
+      },
+      splitBookings.length > 0
```
shmuga: Maybe let’s just return empty array?

src/Services/Air/AirParser.js
```diff
@@ -464,16 +464,49 @@ function extractBookings(obj) {
 
   const travelers = record['common_' + this.uapi_version + ':BookingTraveler'];
   const reservationInfo = record['universal:ProviderReservationInfo'];
+  const remarksObj = record[`common_${this.uapi_version}:GeneralRemark`];
+  const remarks = remarksObj
```
shmuga: Maybe lets also return all remarks?

Issue #266, opened 2017-08-15
AirService.fareRules() request


articice:
Colleagues, for a new front-end functionality request from our partners, I've migrated some old code into uapi-json for a `fareRules` request.

Just like `book`, it receives a list of segments (taken from a previous `shop` reply, **with or without farebasis**) and a long/short param. It performs an AirPriceRequest to look up such flight combination and returns sets of fare rules that apply, associated with each `FareInfoRef` for such a quotation.

Will create PR soon, it will need some code prettify and test coverage. It reuses many validators from existing internal AirPriceReq, and the segments validator from `book`, and we can refactor those using `Joi` which was added as a dependency in the latest version.

Smotrov:
👍

PR #271, opened 2017-08-18
#266 add fareRules() request 

articice:
- Should I rewrite the parser without lodash?
- Should I [uncomment legs grouping which fails on pre-prod](https://github.com/Travelport-Ukraine/uapi-json/commit/6262fb1eb8b133212306eb840aabdfe1dd5ac3ef#diff-1b844efec0b744e58f3f320400e03db2R22) for AirPriceReq, so it works exaclty like one before booking?


articice:
Ready to merge.

src/Services/Air/Air.js
```diff
@@ -15,6 +15,20 @@ module.exports = (settings) => {
       return service.searchLowFares(options);
     },
 
+    fareRules(options) {
```
dchertousov: Do fare rules could be different for different passengers types?
I think this should be an input param.

Please check with Travelport helpdesk before continue.

src/Services/Air/AirParser.js
```diff
@@ -777,6 +777,34 @@ function importRequest(data) {
   return response;
 }
 
+function extractFareRules(obj) {
+  const rulesList = obj['air:FareRule'];
+  _.forEach(rulesList, (item) => {
```
dchertousov: Please use native JS `.forEach` loop.

src/Services/Air/AirParser.js
```diff
@@ -777,6 +777,34 @@ function importRequest(data) {
...
+    const result = [];
+    const listName = (item['air:FareRuleLong']) ? 'air:FareRuleLong' : 'air:FareRuleShort';
+    _.forEach(item[listName], (rule) => {
```
dchertousov: Please use native JS `.forEach` loop.

src/Services/Air/templates/AIR_FARE_RULES_REQUEST.handlebars.js
```diff
@@ -2,21 +2,43 @@ module.exports = `
 <soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/">
     <soapenv:Header/>
     <soapenv:Body>
-        <air:AirFareRulesReq
-            TraceId="test" AuthorizedBy="{{requestId}}" TargetBranch="{{TargetBranch}}"
-            RetrieveProviderReservationDetails="false" FareRuleType="long"
+        <air:AirPriceReq
```
dchertousov: Please describe the need of request type change.

src/Services/Air/Air.js
articice: I've researched this topic before implementing:
Fare rules are linked to a fare basis and route used, and these are the params.

Because fetching fare rules is part of an AirPriceReq, passengers are a required param; passing different passengers creates "virtual" fare quotes for these passengers, applying the fare basis to these passenger types and returning complete **pricing information** about a future, not yet existing booking. In terms of uAPI each **air:AirPricingSolution** / **air:AirPricingInfo** has a **air:FareInfo** calculation; and each **air:FareRule** information is linked to each **air:FareInfo** by reference.

Fare rules text, on the other hand (short and long), **contains information about applying the farebasis to all PTCs** and **are linked to the fare basis itself**. For example, a long description fetched using only ADT, contains text about how this fare is applied to CNN and INF.

We can have different fare bases applied for different passengers (known as \*MULTI\* pricing options as seen in terminal), but currently `fareBasisCode` is a field in `segments`, so basically  if we want to factor out this simplification, we have to move `fareBasisCode` from `segments` into `bookingComponents`, and this is out of scope for current implementation.

There'a s also problem, that sometimes two identical fare information sets are returned, because the same `fareBasisCode` is used for forward and return legs.

dchertousov: > **contains information about applying the farebasis to all PTCs and are linked to the fare basis itself**

That means that if you have fare basis which is available for `CNN` PTC only, then you will not be available to fetch them using `ADT` PTC in request.

src/Services/Air/templates/AIR_FARE_RULES_REQUEST.handlebars.js
articice: this file is not used anywhere, it corresponds to templates.fareRules,
and we're using templates.price for lookupFareRules()

this is a template for booked reservations fare rules request from old code.

articice: it was commited by mistake, will revert it

src/Services/Air/Air.js
articice: That is correct, but the only planned use case for this method was to look up fare rules from segments taken from search results. Because of current search output format limitations (`fareBasisCode` being part of segments, not separate `bookingComponents`) it is impossible to have a search result that uses different fare bases for different PTCs. This is a serious limitation and it should be addressed in another Issue/PR and could lead to change in major version number, because of change in interface.
If a user wants to look up fare rules for an existing booking, I suggest using the native uAPI `FareRulesReq` which is not implemented now (there's an unused template commited in XML templates).

articice: And, you're right, we should change the request format to be analogous to search request.

dchertousov: @articice , please create the ticket to reflect the issue described above.

Also, please, change usage of `_` to native JS functions before continue.

articice: Already did it.
Give me 1 hour to rewrite the request params.

dchertousov:
It's much better with real passengers logic :)
Strange it took so long to fight for it :)

PR #295, opened 2017-11-14
147 retry at ticketing if PNR busy

articice:
implement ticketing retries (with UR import)

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -1,5 +1,6 @@
 import moment from 'moment';
 import _ from 'lodash';
+import retry from 'promise-retry';
 
 import { parsers } from '../../utils';
 import getBookingFromUr from '../../utils/get-booking-from-ur';
@@ -134,19 +135,25 @@ module.exports = (settings) => {
     },
 
     ticket(options) {
-      return this.getPNR(options).then((booking) => {
-        const ticketParams = Object.assign({}, options, {
-          ReservationLocator: booking.uapi_reservation_locator,
-        });
-        return service.ticket(ticketParams)
-          .then(result => result, (err) => {
-            if (err instanceof AirRuntimeError.TicketingFoidRequired) {
-              return this.getPNR(options)
-                .then(updatedBooking => service.foid(updatedBooking))
-                .then(() => service.ticket(ticketParams));
-            }
-            return Promise.reject(err);
+      return retry({ retries: 3 }, (again, number) => {
+        if (settings.debug && number > 1) console.log(`ticket ${options.pnr} retry number ${number}`);
+        return this.getPNR(options).then((booking) => {
+          const ticketParams = Object.assign({}, options, {
+            ReservationLocator: booking.uapi_reservation_locator,
           });
+          return service.ticket(ticketParams)
+            .then(result => result, (err) => {
+              if (err instanceof AirRuntimeError.TicketingFoidRequired) {
+                return this.getPNR(options)
+                  .then(updatedBooking => service.foid(updatedBooking))
+                  .then(() => service.ticket(ticketParams));
+              }
+              if (err instanceof AirRuntimeError.TicketingPNRBusy) {
+                return again(err);
+              }
+              return Promise.reject(err);
+            });
+        });
       });
     },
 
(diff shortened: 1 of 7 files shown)


src/Services/Air/Air.js
```diff
@@ -134,19 +135,28 @@ module.exports = (settings) => {
...
-          .then(result => result, (err) => {
-            if (err instanceof AirRuntimeError.TicketingFoidRequired) {
-              return this.getPNR(options)
-                .then(updatedBooking => service.foid(updatedBooking))
-                .then(() => service.ticket(ticketParams));
-            }
-            return Promise.reject(err);
-          });
+      return retry({ retries: 3 }, (again, number) => {
+        if (settings.debug && number > 1) console.log(`ticket ${options.pnr} retry number ${number}`);
```
dchertousov: Please, use braces for even one statement. It's hard to read and it's prone to error without them.

src/Services/Air/Air.js
```diff
@@ -134,19 +135,28 @@ module.exports = (settings) => {
...
+        return (options.ReservationLocator ? Promise.resolve(options) :
```
dchertousov: Please keep following style:

```
predicate
  ? action
  : another action
```

dchertousov: Also, as you never extend `options` itself, `options` will never have `ReservationLocator` attribute at this point.

src/Services/Air/Air.js
```diff
@@ -134,19 +135,28 @@ module.exports = (settings) => {
...
+          this.getPNR(options).then((booking) => {
+            const ticketParams = Object.assign({}, options, {
+              ReservationLocator: booking.uapi_reservation_locator,
+            });
+            return ticketParams;
+          })
+        ).then(
+          ticketParams => service.ticket(ticketParams)
+            .then(result => result, (err) => {
```
dchertousov: Please, use `catch`, instead of `then` with two params.
If `catch` will have nothing to catch, then result will be returned on its own.

src/Services/Air/Air.js
```diff
@@ -134,19 +135,28 @@ module.exports = (settings) => {
...
+              if (err instanceof AirRuntimeError.TicketingFoidRequired) {
+                return this.getPNR(options)
+                  .then(updatedBooking => service.foid(updatedBooking))
+                  .then(() => this.ticket(ticketParams));
```
dchertousov: Shouldn't be `service.ticket` here?
Otherwise it will be retry in retry.

src/Services/Air/Air.js
articice: Was like it in the original, will rewrite.

dchertousov: Thanks!

src/Services/Air/Air.js
articice: It does have `ReservationLocator` on next retry, see the unit tests added for Air.
It happens because of retry call chain.

src/Services/Air/Air.js
articice: No, I specifically changed it to do a global retry.
Global retry (with optional `ReservationLocator` fetch) instead of an internal one from `service` allows not only simple retry scenarios like PNRBusy->Ticketing, but more complex one like FOID+Busy+Ticketing (depth limited by retry count).

Try changing it into `service` to see a unit test fail.

dchertousov: This gives you more retries, because it's retry inside retry.
I believe this should be changed to:

```javascript
return this.getPNR(options)
  .then(updatedBooking => service.foid(updatedBooking))
  .then(() => again(err));
```

This will make possible for `AirRuntimeError.TicketingFoidRequired` and `AirRuntimeError.TicketingPNRBusy` go one by one.

test/Air/Air.test.js
```diff
@@ -756,6 +756,50 @@ describe('#AirService', () => {
...
+      const getUniversalRecordByPNR = sinon.spy(
+        () => {
+          console.log('UniversalRecord requested');
+          return Promise.resolve(getURbyPNRSampleTicketed);
+        }
+      );
+      const ticketResponses = [
+        Promise.resolve(),
+        Promise.reject(new AirRuntimeError.TicketingPNRBusy([2])),
+        Promise.reject(new AirRuntimeError.TicketingFoidRequired([1])),
```
dchertousov: Test is not correct, because for AirRuntimeError.TicketingFoidRequired response will ALWAYS be `AirRuntimeError.TicketingFoidRequired `

PR #296, opened 2017-11-14
Non-recursive handling proposal

dchertousov:


src/Services/Air/Air.js
```diff
@@ -135,33 +135,31 @@ module.exports = (settings) => {
...
-            : this.getPNR(options).then((booking) => {
-              const ticketParams = Object.assign({}, options, {
-                ReservationLocator: booking.uapi_reservation_locator,
-              });
-              return ticketParams;
-            })
-        ).then(
-          ticketParams => service.ticket(ticketParams)
+      return (options.ReservationLocator
+        ? Promise.resolve(options.ReservationLocator)
```
dchertousov: @articice I have coverage issue here.
Could the `ReservationLocator` be directly passed on the first call?

PR #301, opened 2017-11-16
Added stopover to coupons.

shmuga:


dchertousov:
Please, update version, if it's the only pending enhancement.

PR #315, opened 2018-01-16
Added availability request with tests and docs.

shmuga:


src/Services/Air/AirParser.js
```diff
@@ -994,6 +994,60 @@ function exchangeBooking(rsp) {
...
+
+    leg.push(s);
+
+    if (!isConnected) {
+      results.push(leg);
+      leg = [];
+    }
+  });
+
+  if (leg.length !== 0) {
```
dchertousov: When `leg.length` could equal `0`?
Is it an error to throw?

src/Services/Air/AirParser.js
```diff
@@ -994,6 +994,60 @@ function exchangeBooking(rsp) {
...
+    results.push(leg);
+  }
+
+  return {
+    legs: results,
+    nextResultReference: rsp[`common_${this.uapi_version}:NextResultReference`],
```
dchertousov: You return reference but you never use it in requests. Shouldn't it be added as a param?

src/Services/Air/AirParser.js
shmuga: No, It can be caused when last segment is connection.

src/Services/Air/AirParser.js
shmuga: Yes, I missed this.

PR #318, opened 2018-01-26
Added possibility to return all classes.

shmuga:
Added status codes for availability.


src/Services/Air/AirParser.js
```diff
@@ -1009,7 +1009,15 @@ function availability(rsp) {
...
-    const cabinsAvailability = availInfo
+    if (!availInfo) {
+      return;
+    }
+
+    if (!availInfo['air:BookingCodeInfo']) {
+      return;
+    }
+
+    const cabinsAvailability = this.env.cabins && this.env.cabins.length > 0
```
dchertousov: I'd really like to see this filed calculation simplified.

PR #326, opened 2018-04-02
Added pcc to errors NoAgreement

shmuga:


dchertousov:
@shmuga , please update version before merge

PR #327, opened 2018-04-12
Hotfix for pcc parsing in no  agreement error.

shmuga:


src/utils/get-error-pcc.js
```diff
@@ -3,7 +3,7 @@ export default (string) => {
     return null;
   }
 
-  const [, pcc = null] = string.match(/NO AGREEMENT EXISTS FOR AGENCY - ([A-Z0-9]+)/) || [];
+  const [, pcc = null] = string.match(/NO AGREEMENT EXISTS FOR AGENCY(?:\s+)-(?:\s+)([A-Z0-9]+)/) || [];
```
dchertousov: Could be simplified to `\s+-\s+`.
Please check and update if it works.

Also I would like to see PCC limited to 3-4 symbols:
`([A-Z0-9]{3,4})`

PR #346, opened 2018-10-31
PassiveRemark without service segment

cursedcoder:
@dchertousov please suggest if it is correct to handle remark in such way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "1.0.0",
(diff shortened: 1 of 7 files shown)


dchertousov:
Well, there are more different types of service segments, which are not remarks.
But I would prefer to go along with this code until we need to dive deeper.

dchertousov:
If you don't have any more changes before release, please update the version, and let's release it.

dchertousov:
Please make npm i before push, to include version bump into package-lock.json

Issue #344, opened 2018-10-18
cancelPNR enhancement


dchertousov:
When tickets in PNR are not voided, but refunded, and no `cancelTickets` is passed PNR should be canceled anyway

dchertousov:
Closed by #351

PR #351, opened 2018-11-19
issue/344 | Cancel PRN 

InkFaust:
#344

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -296,6 +296,15 @@ module.exports = (settings) => {
               if (allTicketsVoid) {
                 return Promise.resolve(true);
               }
+              // Check for REFUNDED
+              const allTicketsRefunded = ticketData.tickets.every(
+                ticket => ticket.coupons.every(
+                  coupon => coupon.status === 'R'
+                )
+              );
+              if (!allTicketsVoid && allTicketsRefunded && !options.cancelTickets) {
+                return Promise.resolve(true);
+              }
               // Check for cancelTicket option
               if (options.cancelTickets !== true) {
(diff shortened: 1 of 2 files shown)


src/Services/Air/Air.js
```diff
@@ -296,6 +296,15 @@ module.exports = (settings) => {
               if (allTicketsVoid) {
                 return Promise.resolve(true);
               }
+              // Check for REFUNDED
+              const allTicketsRefunded = ticketData.tickets.every(
+                ticket => ticket.coupons.every(
+                  coupon => coupon.status === 'R'
+                )
+              );
+              if (!allTicketsVoid && allTicketsRefunded && !options.cancelTickets) {
```
cursedcoder: first check is redundant

test/Air/Air.test.js
```diff
@@ -1394,6 +1394,45 @@ describe('#AirService', () => {
           expect(cancelPNR).to.have.callCount(0);
         });
     });
+    it('should succeed when there are no VOID, but REFUNDED tickets and cancelTickets = false', () => {
```
cursedcoder: not exactly cancelTickets = false,
but just when `there is no cancelTickets`

src/Services/Air/Air.js
```diff
@@ -296,6 +296,15 @@ module.exports = (settings) => {
               if (allTicketsVoid) {
                 return Promise.resolve(true);
               }
+              // Check for REFUNDED
+              const allTicketsRefunded = ticketData.tickets.every(
```
dchertousov: There is no need in checking array twice. All tickets should be either VOID or REFUND

Issue #93, opened 2017-01-13
Make TAU timestamp configurable


dchertousov:
In order to get `book` method of `AirService` working TAU modifier is set to 3 hours from the current timestamp.

Is there a need to let user to configure this settings during the request?

@Smotrov , please advise.

PR #354, opened 2018-11-20
Issue/93 Configurable TAU

InkFaust:
## PR Purpose
Assigned issue #93.
- Implement configurable TAU option for .book() method of Air service
- Aid docs upd
- Air book test upd

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -36,8 +36,11 @@ module.exports = (settings) => {
 
     book(options) {
       return service.airPricePricingSolutionXML(options).then((data) => {
+        const tau = options.tau && moment(options.tau).isValid()
+          ? moment(options.tau).format()
+          : moment().add(3, 'hours').format();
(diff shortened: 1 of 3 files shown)


src/Services/Air/Air.js
```diff
@@ -36,8 +36,11 @@ module.exports = (settings) => {
 
     book(options) {
       return service.airPricePricingSolutionXML(options).then((data) => {
+        const tau = options.tau && moment(options.tau).isValid()
```
cursedcoder: Will this work also?
```
const tauDate = moment(options.tau);
const tau = tauDate.isValid() ? tauDate.format() : moment().add(3, 'hours').format();
```

InkFaust: But this will:
```javascript
const tauDate = moment(options.tau || null);
const tau = tauDate.isValid() ? tauDate.format() : moment().add(3, 'hours').format();
```

Issue #313, opened 2018-01-02
Preferred Provider: 1G is not in the permissible list of providers for this user and operation.


joshdhenry:
I am trying the below request but I get the error "Preferred Provider: 1G is not in the permissible list of providers for this user and operation."  I think my preferred provider needs to be 1P. Is there any way to change the preferred provider?

```
const settings = 
    { 
        auth: {
            username: 'my username', 
            password: 'my password', 
            targetBranch: 'my target branch',
        },
        production: false,
        debug: 2,
    };

const AirService = uAPI.createAirService(settings);

const params = {
  legs: [
    {
      from: 'LWO',
      to: 'JKT',
      departureDate: '2018-07-18'
    },
    {
      from: 'JKT',
      to: 'IEV',
      departureDate: '2018-07-21',
    },
  ],
  passengers: {
    ADT: 1,
  },
  cabins: ['Economy'], 
  requestId: '4e2fd1f8-2221-4b6c-bb6e-cf05c367cf60',
  pricing: {
    currency: 'USD',
  },
};

AirService.shop(params)
  .then(
    res => console.log(res),
    err => console.log(err)
  );
```

dchertousov:
Hello!

The library was never tested in 1P, as we never had such tasks and access to 1P PCCs.

Feel free to fork, build and try it by yourself.

You may need to run the test script with Babel-node (see babel-cli).

Comments about 1P compatibility will be most appreciated!

joshdhenry:
I will fork it and see how it goes.  Thanks for the response and for this great tool!

dchertousov:
@joshdhenry have you tried it with 1P? Did you have any issues?
Do you think, you may extend configuration to include provider?

joshdhenry:
@dchertousov I was able to get some basic air requests working by changing all instances of the string "1G" to "1P". There may have been a bit more to it, but it wasn't that difficult to get it working.  I didn't end up using this in my project, though, so you may see issues I didn't notice.

dchertousov:
@joshdhenry Thank you for the response! Could I ask you what library did you finally used and why have you chosen them over uapi-json?

dchertousov:
@InkFaust please investigate the issue and prepare your questions.
I think we should make provider configurable with default value of `1G`

InkFaust:
@dchertousov got it.
I think configurable param for provider selection should be declared within `auth` object as optional param with `1G` default value.

dchertousov:
[...]

PR #355, opened 2018-11-23
issue/313 | Define configurable provider param of auth

InkFaust:
# Purpose
According to ticket #313, add configurable provider parameter

### DIFF
--- a/src/Request/uapi-request.js
+++ b/src/Request/uapi-request.js
@@ -172,10 +172,7 @@ module.exports = function uapiRequest(
       .then(sendRequest)
       .then(parseResponse)
       .then(validateSOAP)
-      .then((res) => {
-        const parser = parseFunction.bind(uParser);
-        return parser(res);
-      }) // TODO: merge Hotels
+      .then(parseFunction.bind(uParser)) // TODO: merge Hotels
       .then(handleSuccess);
   };
 };
--- a/src/Request/uapi-request.js
+++ b/src/Request/uapi-request.js
@@ -37,19 +37,24 @@ module.exports = function uapiRequest(
   debugMode = false,
   options = {}
 ) {
-  const config = configInit(auth.region || 'emea');
+  // Assign default value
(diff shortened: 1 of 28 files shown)


cursedcoder:
Also, please bump version to the next minor

dchertousov:
@InkFaust , please bump version to the 1.0.3

src/Request/uapi-request.js
```diff
@@ -45,6 +45,8 @@ module.exports = function uapiRequest(
     throw new RequestValidationError.ServiceUrlMissing();
   } else if (!auth || auth.username === undefined || auth.password === undefined) {
     throw new RequestValidationError.AuthDataMissing();
+  } else if (!auth.provider) {
+    auth.provider = '1G';
```
dchertousov: We are checking for errors here.
So if provider will not be set, you will set it here, but this way you will stop checking for errors.

Please move it out of this if/else block.
And it would be even better to refactor this if/else block to separate if blocks

src/Request/uapi-request.js
```diff
@@ -164,7 +167,12 @@ module.exports = function uapiRequest(
       .then(sendRequest)
       .then(parseResponse)
       .then(validateSOAP)
-      .then(parseFunction.bind(uParser))// TODO merge Hotels
+      .then((res) => {
+        const parser = parseFunction.bind(uParser);
+        return parser(res, {
```
dchertousov: You have already passed auth.provider in this parser in line 62, why do you need to pass it once again, and can't handle provider in the parser itself?

src/Request/uapi-request.js
```diff
@@ -169,9 +174,7 @@ module.exports = function uapiRequest(
       .then(validateSOAP)
       .then((res) => {
```
dchertousov: We could return to the original (if it works with provider passed like above):
```
.then(parseFunction.bind(uParser))
```

InkFaust: Indeed!

Issue #359, opened 2018-12-20
Canceling PNR without tickets


Dezaimasu:
When trying to call cancelPNR method for recently created PNR, without any issued tickets, error is returned.

It happens because cancelPNR tries to get tickets and check their status before canceling PNR, and getTickets methods throws exception if there's no tickets 
https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/Air.js#L206

Argument "cancelTickets" for cancelPNR changes nothing. It tries to retrieve tickets either way.

Would it be possible to call for this:

https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/Air.js#L329-L333

skipping this whole part that checks tickets' current state: 

https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/Air.js#L286-L328

dchertousov:
Hello @Dezaimasu !

It's so nice you dig so deep into the code!

You could open your own pull request and we will incorporate your changes, or we will try to fix this issue later next week.

Dezaimasu:
I think there's a bug in the code, when you're calling getUniversalRecordByPNR inside cancelPNR, it returns not a record, but an array of records, so "record.uapi_reservation_locator" wouldn't find anything. 

https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/Air.js#L286-L288

Tried to fix that, but now tests are failing. I'll add it to pull request anyway because I'm quite sure it was an actual bug and tests should be adjusted as well. Just reject it if I'm wrong.

dchertousov:
@Dezaimasu , good!
Could you write test for it and create a PR?

dchertousov:
Closed in #360

PR #360, opened 2018-12-22
Issue 359. Canceling PNR without tickets

Dezaimasu:
Now cancelPNR function won't try to get tickets in advance if getUniversalRecordByPNR returned PNR without any tickets.
There's some bug mentioned in commentary here:
https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/Air.js#L69-L70
If getUniversalRecordByPNR  might return PNR without tickets in some other cases as well, record.tickets.length check should be replaced with something else. 

This check is probably not required and could be removed, because getTickets theoretically should just return an empty list if there was no tickets in requested PNR, but it crashes inside AirErrorHandler on this place for me:
https://github.com/Travelport-Ukraine/uapi-json/blob/abfc3080c078e75c607956c7d05e1a022db7ded9/src/Services/Air/AirParser.js#L306-L307
It tries to call map on undefined, there's no air:AirSegmentError element in my errorInfo:

> {
> 	 "common_v39_0:Code": "3000",
> 	"common_v39_0:Service": "PROVISIONINGSERVICE",
> 	"common_v39_0:Type": "",
> 	"common_v39_0:Description": "The reservation 'AY91L8' has no tickets yet",
> 	"common_v39_0:TransactionId": "D62913960A07643C89F1F5063434A5A3",
> 	"xmlns:common_v39_0": "http://www.travelport.com/schema/common_v39_0"
> }
Probably some non-standard response from Travelport uAPI.

And I fixed the bug I mentioned in https://github.com/Travelport-Ukraine/uapi-json/issues/359#issuecomment-449575377 Unfortunately I'm not very familiar with tests, so I'm leaving them to someone else. Sorry about that.

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -283,54 +283,58 @@ module.exports = (settings) => {
     },
 
     cancelPNR(options) {
+      const ignoreTickets = typeof options.ignoreTickets === 'undefined'
+        ? false // default value
+        : options.ignoreTickets;
+
+      const checkTickets = (tickets) => {
+        return Promise.all(tickets.map(
+          (ticketData) => {
+            // Check for VOID or REFUND
+            const allTicketsVoidOrRefund = ticketData.tickets.every(
+              ticket => ticket.coupons.every(
+                coupon => coupon.status === 'V' || coupon.status === 'R'
+              )
+            );
+            if (allTicketsVoidOrRefund) {
+              return Promise.resolve(true);
+            }
+            // Check for cancelTicket option
+            if (options.cancelTickets !== true) {
+              return Promise.reject(new AirRuntimeError.PNRHasOpenTickets());
+            }
+            // Check for not OPEN/VOID segments
+            const hasNotOpenSegment = ticketData.tickets.some(
+              ticket => ticket.coupons.some(
+                coupon => 'OV'.indexOf(coupon.status) === -1
+              )
+            );
+            if (hasNotOpenSegment) {
+              return Promise.reject(new AirRuntimeError.UnableToCancelTicketStatusNotOpen());
+            }
+            return Promise.all(
+              ticketData.tickets.map(
+                ticket => (
+                  ticket.coupons[0].status !== 'V'
+                    ? service.cancelTicket({
+                      pnr: options.pnr,
+                      ticketNumber: ticket.ticketNumber,
+                    })
+                    : Promise.resolve(true)
+                )
+              )
+            );
+          }
+        ));
+      };
+
       return this.getUniversalRecordByPNR(options)
         .then((ur) => {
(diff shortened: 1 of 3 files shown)


dchertousov:
@Dezaimasu you have some issues with your code, as build does not pass. Please see build details

src/Services/Air/Air.js
```diff
@@ -283,54 +283,66 @@ module.exports = (settings) => {
...
-                        pnr: options.pnr,
-                        ticketNumber: ticket.ticketNumber,
-                      })
-                      : Promise.resolve(true)
-                  )
-                )
-              );
-            }
-          ))
+          ur => getBookingFromUr(ur, options.pnr)
```
dchertousov: I recommend you to use `this.getTickets(options)` instead, as tickets from import call could be corrupted. That's why tickets are filtered in response.

src/Services/Air/Air.js
```diff
@@ -284,21 +284,15 @@ module.exports = (settings) => {
...
-              ? Promise.resolve([])
-              : this.getTickets({ reservationLocatorCode: record.uapi_reservation_locator })
-          )
-            .then(
-              tickets => Promise.all(tickets.map(
+        .then((ur) => {
+          const record = Array.isArray(ur) ? ur[0] : ur;
+
+          return this.getTickets(record)
+            .catch(() => {
```
dchertousov: If the ticket couldn't be retrieved you will get an error message like this.

```xml
<SOAP:Envelope 
  xmlns:SOAP="http://schemas.xmlsoap.org/soap/envelope/">
  <SOAP:Body>
    <air:AirRetrieveDocumentRsp TransactionId="F54E52500A07405C5494698BE234E2CA" ResponseTime="90" 
      xmlns:air="http://www.travelport.com/schema/air_v39_0" 
      xmlns:common_v39_0="http://www.travelport.com/schema/common_v39_0">
      <common_v39_0:ResponseMessage Code="12009" Type="Error" ProviderCode="1G">Host error during ticket retrieve.- Ticket Number - UNABLE TO PROCESS INVALID TICKET NUMBER</common_v39_0:ResponseMessage>
    </air:AirRetrieveDocumentRsp>
  </SOAP:Body>
</SOAP:Envelope>
```
This will generate error for you.
Sometimes you can get error because of uAPI outage.

If we just return empty array here we could cancel segments with still active ticket.

Either remove this check with returning empty array, or add additional param `ignoreTickets=false|true` to the `cancelPNR` method to skip tickets check. This way you can skip the check in case of error.

dchertousov:
New version has been published to npm.
In future we plan to return empty array instead of uAPI error in case when UR contains no tickets.

PR #367, opened 2019-02-04
provide PlatingCarrier AirPricingModifier

paulvoloschuk:


test/Air/validators/platingCarrier-optional.test.js
```diff
@@ -0,0 +1,16 @@
...
+
+const validate = require('../../../src/Services/Air/validators/platingCarrier');
+
+describe('Air.validators.platingCarrier', () => {
+  it('should pass', () => {
+    const platingCarrier = 'A1';
+    validate({ platingCarrier });
+  });
+
+  it('should throw error', () => {
```
dchertousov: Add test for undefined valued, e.g. validate `{}`

src/Services/Air/validators/platingCarrier.js
```diff
@@ -0,0 +1,7 @@
+const { AirValidationError } = require('../AirErrors');
+
+module.exports = (params) => {
+  if (!String(params.platingCarrier).match(/^[A-Z0-9]{2}$/)) {
```
dchertousov: Should be optional

test/Air/validators/platingCarrier-optional.test.js
dchertousov: XML format should be covered also

PR #370, opened 2019-02-07
Issue/362

paulvoloschuk:
`.getTickets()` will return empty if UR has no tickets

### DIFF
--- a/test/Air/Air.test.js
+++ b/test/Air/Air.test.js
@@ -1041,7 +1041,7 @@ describe('#AirService', () => {
       });
       const service = createAirService({ auth });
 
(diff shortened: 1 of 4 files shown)


test/Air/Air.test.js
```diff
@@ -1041,7 +1041,7 @@ describe('#AirService', () => {
       });
       const service = createAirService({ auth });
 
-      service.getTickets({ uapi_reservation_locator: 'RLC001' })
+       service.getTickets({ uapi_reservation_locator: 'RLC001' })
```
dchertousov: Could you please align spaces here?

PR #375, opened 2019-03-05
issue/374 | Include terminal id into terminal logs

InkFaust:
I've decided to not include terminal id into the state since it has no reason to be there at the moment of time due to the flow of using the terminal id in one scope.


src/Services/Terminal/Terminal.js
```diff
@@ -91,43 +92,39 @@ module.exports = function (settings) {
       .then(resolve)
       .catch(reject);
   });
+  // Get terminal ID
+  const getTerminalId = sessionToken => Promise.resolve(getHashSubstr(sessionToken));
```
dchertousov: It's sync function

src/Services/Terminal/Terminal.js
```diff
@@ -91,43 +92,39 @@ module.exports = function (settings) {
...
 
   const terminal = {
     getToken: getSessionToken,
-    executeCommand: (command, stopMD = defaultStopMD) => Promise.resolve()
-      .then(() => {
+    executeCommand: (command, stopMD = defaultStopMD) => new Promise(async (resolve) => {
+      try {
+        const sessionToken = await getSessionToken();
+        const terminalId = await getTerminalId(sessionToken);
```
dchertousov: Use function as sync

PR #392, opened 2019-04-23
Typo bug fix in  air price  req template

engrashid:
there was a typo in  {{#if plattingCarrier}}  
corrected to
{{#if platingCarrier}}


dchertousov:
Thank you @engrashid!
We are going to release this fix as a part of 1.1.2 release

PR #397, opened 2019-05-04
getTickets update

dchertousov:
- Fixed hadnling of PNRs without tickets.
- geTtickets now always returns array
- Dependencies updated to maintain package security

### DIFF
--- a/src/Services/Air/AirParser.js
+++ b/src/Services/Air/AirParser.js
@@ -505,17 +505,6 @@ const airGetTicket = function (obj) {
     throw new AirRuntimeError.TicketRetrieveError(obj);
   }
 
-  if (
-    obj['SOAP:Fault']
-    && obj['SOAP:Fault'][0]
-    && obj['SOAP:Fault'][0].detail[0]
-    && obj['SOAP:Fault'][0].detail[0][`common_${this.uapi_version}:ErrorInfo`][0]
-    && obj['SOAP:Fault'][0].detail[0][`common_${this.uapi_version}:ErrorInfo`][0][`common_${this.uapi_version}:Code`][0]
-    === '3000'
-  ) {
-    return [];
-  }
-
   const etr = obj['air:ETR'];
   if (!etr) {
     throw new AirRuntimeError.TicketRetrieveError(obj);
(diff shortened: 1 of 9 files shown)


src/Services/Air/AirParser.js
```diff
@@ -531,6 +520,55 @@ const airGetTicket = function (obj) {
   return getTicketFromEtr.call(this, etr, obj);
 };
 
+function airGetTicketsErrorHandler(rsp) {
+  let errorInfo;
+  let code;
+  try {
+    errorInfo = (
```
InkFaust: What will happen later on if the condition returns no data? Or it's impossible to have no data returned from the response in this case? If it's possible, maybe it's better to prevent any possible errors that could be occurred on L531

src/Services/Air/AirParser.js
```diff
@@ -531,6 +520,55 @@ const airGetTicket = function (obj) {
...
+    code = errorInfo[`common_${this.uapi_version}:Code`];
+  } catch (err) {
+    throw new RequestRuntimeError.UnhandledError(null, new AirRuntimeError(rsp));
+  }
+  // General Air Service error
+  if (code === '3000') {
+    if (
+      rsp.faultcode !== undefined
+      && rsp.faultstring !== undefined
+      && /has no tickets/.test(rsp.faultstring) === true
```
InkFaust: A regexp test will return boolean `true` in case if it's passed and `false` in case if it's not, I don't think it's necessary to make a check if the value is true or not.

src/Services/Air/AirParser.js
```diff
@@ -531,6 +520,55 @@ const airGetTicket = function (obj) {
...
+    ) {
+      return rsp;
+    }
+  }
+  switch (code) {
+    case '345':
```
InkFaust: If it's the only one case then maybe `if/else` will be better?

InkFaust: Unless it's possible to have more cases in the future.

src/Services/Air/AirParser.js
```diff
@@ -531,6 +520,55 @@ const airGetTicket = function (obj) {
...
+      throw new RequestRuntimeError.UnhandledError(null, new AirRuntimeError(rsp));
+  }
+}
+
+function airGetTickets(obj) {
+  // No tickets in PNR
+  if (
+    obj.faultcode !== undefined
+    && obj.faultstring !== undefined
+    && /has no tickets/.test(obj.faultstring) === true
```
InkFaust: Same as above.
Also, the condition is repeated twice, I think could be reused.

src/Services/Air/AirParser.js
```diff
@@ -531,6 +520,55 @@ const airGetTicket = function (obj) {
...
+  ) {
+    return [];
+  }
+  // Parsing response
+  const tickets = airGetTicket.call(this, obj);
+  if (Array.isArray(tickets)) {
```
InkFaust: Maybe it's better to declare the return value by using conditional operator rather than `if` statement. At least it will take less space.

src/Services/Air/AirParser.js
dchertousov: Then code will fail and RequestRuntimeError.UnhandledError will be thrown. It's ok here.

src/Services/Air/AirParser.js
dchertousov: More cases could be added.

src/Services/Air/AirParser.js
InkFaust: Got it, thank you.

PR #398, opened 2019-05-06
Air ticket req template - Adds airprice info node to airticketingmodifier

engrashid:
AirPriceReq info node is necessary to book ticket that has child passenger . 
And it has to be included inside AirTicketingModifiers in air ticket request.

### DIFF
--- a/src/Services/Air/templates/AIR_TICKET_REQUEST.handlebars.js
+++ b/src/Services/Air/templates/AIR_TICKET_REQUEST.handlebars.js
@@ -16,6 +16,9 @@ module.exports = `
       <air:AirPricingInfoRef Key="{{key}}"/>
       {{/refs}}
       <air:AirTicketingModifiers>
+        {{#refs}}
+          <air:AirPricingInfoRef Key="{{key}}"/>
+        {{/refs}}
         {{#if commission}}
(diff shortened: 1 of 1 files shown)


src/Services/Air/templates/AIR_TICKET_REQUEST.handlebars.js
```diff
@@ -16,6 +16,9 @@ module.exports = `
       <air:AirPricingInfoRef Key="{{key}}"/>
       {{/refs}}
       <air:AirTicketingModifiers>
+        {{#refs}}
```
dchertousov: I see similar construction on the line 15.
Should the new lines you propose be added, or moved here from there?

engrashid: New lines should be added . If we add child in ticketing then it becomes mandatory to include refs inside AirTicketingModifiers . So both will stay .But we could think of something that will put a condition if any child passenger only then added part will go there.

dchertousov: @engrashid it's ok as it is, if it's not breaking anything

PR #399, opened 2019-05-13
Add separate AirPrice call so that LFS results can be confirmed and a…

creduo:
this commit will takes segment, passengers for arguments and will return similar to LFS search result

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -20,6 +20,10 @@ module.exports = (settings) => {
       return service.availability(options);
     },
 
+    airPrice(options) {
+      return service.airPrice(options);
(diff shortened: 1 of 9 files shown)


test/testconfig.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  username: 'USERNAME',
-  password: 'PASSWORD',
-  targetBranch: 'BRANCH',
+  username: 'Universal API/uAPI5325526267-a07def41',
```
dchertousov: I'm not sure you want to share your credentials with the community

creduo: Yep this was mistaken.

Issue #400, opened 2019-05-13
Correct condition of searchLowFaresValidate() in AirParser.js


creduo:
https://github.com/Travelport-Ukraine/uapi-json/blob/902bcb817bee457780d048395dc5e5d31bba3773/src/Services/Air/AirParser.js#L35

Condition is always false since types 'boolean' and 'string' have no overlap

dchertousov:
Done in #401

PR #401, opened 2019-05-13
Condition is always false since types 'boolean' and 'string' have no …

creduo:
changed 
if (!Object.prototype.toString.call(obj[airName]) === '[object Object]') 
to 
if (Object.prototype.toString.call(obj[airName]) !== '[object Object]')

this PR might result AirParsingError.ResponseDataMissing to someone using SolutionResult in LFS


creduo:
Issue #400

Issue #403, opened 2019-05-14
Support for solutionResults in Low Fare Search Request


creduo:
uAPI LFS returns two kinds of results, AirPricePointList and AirPricingSolution but formatLowFaresSearch only parses AirPricePointList.

It'd be good to have AirPricingSolution support (SolutionResults) too.

Smotrov:
Hi @creduo! 
Yes, you are right. 
Unfortunately, Shopping functions are not in our current priority.
However but though, we would just love to accept your contribution if you feel you can cope with that.

PR #404, opened 2019-05-14
add AirPricingSolution support for LFS

creduo:
Please let me know if this approach is applicable to this project to solve #403


src/Services/Air/AirFormat.js
```diff
@@ -283,57 +284,103 @@ function formatLowFaresSearch(searchRequest, searchResult) {
 
   const fares = [];
 
-  Object.entries(pricesList).forEach(([fareKey, price]) => {
+  const isSolutionResult = typeof solutionsList !== 'undefined';
+
+  const results = isSolutionResult ? solutionsList : pricesList;
```
cursedcoder: Is that a BC change? If so, please bump the minor version in pkg.json
Not pretty sure what's the difference between solution/price, could you clarify that in the docs?

creduo: solution returns only one schedule combination per pricing option per item, which is used in common nowadays, while AirPricePoint returns all available schedule per pricing options.

* [AirPricingSolutions](https://support.travelport.com/webhelp/uapi/Content/Air/Low_Fare_Shopping/Low_Fare_Shopping_(Synchronous).htm#AirPricingSolutions)

* [AirPricePointList]( https://support.travelport.com/webhelp/uapi/Content/Air/Low_Fare_Shopping/Low_Fare_Shopping_by_Price_Points.htm)

dchertousov:
Approved, but improved code coverage will be appreciated

Issue #321, opened 2018-02-13
Query: Async Low Fare Search request & Support for 1G


skills-up:
Hi,

Just wish to know whether the API is using Async version of LowFareSearch request, or the synchronized one? Also, going by the issue log, there seems to be a lack of support for Provider = '1G'. Is this the case, or an isolated issue?

Apologies for posting the query as issue, in the absence of any direct mechanism for posting these questions.

shmuga:
Hello @skills-up 

For LFS we are using sync version. You can check that [here](https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Services/Air/templates/AIR_LOW_FARE_SEARCH_REQUEST.handlebars.js) 
For now we are supporting only 1G provider. There is no support of any other.

skills-up:
Just curious, in your opinion, would async version provide a performance improvement? Specially for listing options, and booking a selected option among the retrieved list.

articice:
@skills-up there only will be an improvement of resource usage on your side and on the front-end in the simplest use case. With async request, your server will have to process a smaller amount of options, and you can add paging or infinite scroll to your website (which preserves resources). Applicability of such solution depends on your desired functionality and system design.

We've never tested it, but IMHO the initial search time is not likely to improve. There's a basic undividable search operation underneath with request time of around 5-15 sec, which creates a TTFB (time to first byte) and it should be the same for async. It also doesn't depend on your level of provisioning, e.g. level 3 generates hundreds of pricing options (up to completely absurd ones), but search time is not longer compared to previous level.

If you're not quite satisfied with such long response times, you should try:
- provisioning e-Pricing meta search from Travelport (a simpler and faster cached request, caching provided by Travelport)
- connect an external cache, e.g. https://www.cee-systems.com/estreamingapi
- create your own caching solution.

skills-up:
The reason I was looking for an improvement is because I've to query two fares: public fares and private fares, and then show them side by side. Since there is no option to fetch them in a single request, we end up making two requests, and the response time accordingly double to an average 13 seconds. This is basically a huge amount to time from the end user perspective. Thus, I was looking for speeding up things a bit.

PR #406, opened 2019-05-14
Added LFS Async and Retrieve method. options.async = true

creduo:
issue #321 

optional param async = true to trigger LowFareSearchAsynch
preproduction does not work with async request currently,
will return almost same result in same execution time with LowFareSearchReq.


src/Services/Air/Air.js
```diff
@@ -13,9 +13,16 @@ module.exports = (settings) => {
   const log = (settings.options && settings.options.logFunction) || console.log;
   return {
     shop(options) {
+      if (typeof options.async !== 'undefined' && options.async === true) {
```
dchertousov: Check for undefined is not needed as we compare to true with strong type matching

src/Services/Air/templates/AIR_LOW_FARE_SEARCH_ASYNCH_REQUEST.handlebars.js
```diff
@@ -0,0 +1,105 @@
+module.exports = `
```
dchertousov: Please fix `asynch` in file name

dchertousov: Does this template have a lot of changes comparing to regular LFS?
Is there any need to keep two version of them? IMO it's nice to have one and include some blocks based on request type.

src/Services/Air/Air.js
```diff
@@ -13,9 +13,16 @@ module.exports = (settings) => {
   const log = (settings.options && settings.options.logFunction) || console.log;
   return {
     shop(options) {
+      if (options.async === true) {
+        return service.searchLowFaresAsync(options);
+      }
       return service.searchLowFares(options);
     },
 
+    retrieve(options) {
```
shmuga: Maybe it will make sense to be more specific with retrieve function, like `retrieveShop`? Otherwise, it may cause confusion with retrieve PNR

src/Services/Air/templates/AIR_LOW_FARE_SEARCH_REQUEST.handlebars.js
```diff
@@ -4,12 +4,21 @@ module.exports = `
 <!--Air Low Fare Search For Galileo({{provider}}) Request-->
 <soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
     <soap:Body>
+        {{#if async}}
+        <air:LowFareSearchAsynchReq
+            AuthorizedBy="user" TraceId="{{requestId}}" TargetBranch="{{TargetBranch}}"
```
shmuga: This part seems duplicated below. Isn't enough just to change an opening tag in case of the async method?

src/Services/Air/AirService.js
```diff
@@ -22,6 +22,28 @@ module.exports = function (settings) {
...
       options
     ),
+    searchLowFaresAsync: uApiRequest(
+      config.AirService.url,
+      auth,
+      templates.lowFareSearch,
+      'air:LowFareSearchAsynchRsp',
+      AirValidator.AIR_LOW_FARE_SEARCH_REQUEST,
+      AirParser.AIR_ERRORS,
+      AirParser.AIR_LOW_FARE_SEARCH_REQUEST,
```
shmuga: What will return this parser in case of async request? Will it be possible to get `searchId` to use in `retrieveShop` function?

src/Services/Air/templates/AIR_LOW_FARE_SEARCH_REQUEST.handlebars.js
creduo: Maybe, but wouldn't it be more clear to understand  if they're in separate lines?

Do you prefer 
```
<air:{{#if async}}LowFareSearchAsynchReq{{#else}}LowFareSearchReq{{/if}}
```

rather than this?
```
{{#if async}}
<air:LowFareSearchAsynchReq ...
{{#else}}
<air:LowFareSearchReq ...
{{/if}}
```

src/Services/Air/Air.js
creduo: Yes, I agree, there're too many retrieves like retrieveUR, retrieveTicket etc, 
I renamed the retrieve() with retrieveShop as you mentioned

src/Services/Air/AirService.js
creduo: Because the formatLowFaresSearch in AirFormat.js only returns list of fares, it is not possible to return the searchId without breaking the current structures, should I change the code to return optional like `return options.faresOnly ? response.fares : response;` ?

src/Services/Air/templates/AIR_LOW_FARE_SEARCH_ASYNCH_REQUEST.handlebars.js
creduo: Commit 1a5ffed merged async template to lfs template, no more file named asynch.

src/Services/Air/Air.js
creduo: Changed as commented.

Issue #408, opened 2019-05-17
Add UniversalRecordRetrieve method


creduo:
Currently UniversalRecordImport is implemented and is being used to retrieve UR.

But separate UniversalRecordRetrieve call may be needed for certain workflows (for ex. doesn't want to import PNR if not exists etc.)

creduo:
Added PR #409 for this feature :D

creduo:
PR was merged and included in release 1.2.2

PR #409, opened 2019-05-20
Add UniversalRecordRetrieve method

creduo:


creduo:
This PR is for issue #408

src/Services/Air/AirParser.js
```diff
@@ -925,6 +925,11 @@ function importRequest(data) {
   return response;
 }
 
+function universalRecordRetrieveRequest(data) {
+  const response = extractBookings.call(this, data);
```
shmuga: Have you tried calling this request with booking which was not imported before retrieval? What will be the behaviour of this parser?

creduo: It will raise RuntimeError.AirRuntimeError "Record locator not found", which I thought should be adequate, should it return an empty array?

docs/Air.md
```diff
@@ -238,6 +239,21 @@ If the PNR contains no active segments it could not be imported into uAPI. Thus
 
 **See: <a href="../examples/Air/getUniversalRecordByPNR.js">Example</a>**
 
+## .getUniversalRecord(params)
+<a name="getUniversalRecord"></a>
+> May require Terminal access enabled in uAPI. See [TerminalService](Terminal.md)
```
shmuga: It seems that this method does not use terminal service

creduo: My bad, I just copied from above.

PR #415, opened 2019-07-10
[NPM] Updates packages

shmuga:
### Scope
Some of the packages are marked as deprecated and vulnerable.
Using npm-check-updates new versions are set. Eslint, chai, sinon
are left unchanged due to incompatible changes.

### Changes
- Updates package.json with lock
- Changes joi to @hapi/joi


cursedcoder:
Joi seems to be used only in 1 place to validate credit card – that's odd :)

shmuga:
> Joi seems to be used only in 1 place to validate credit card – that's odd :)

Yep, I know. But it does it job and validates a card 🙆‍♂

dchertousov:
Created issue to replace joi: #420

PR #418, opened 2019-07-11
Booking with warnings

cursedcoder:


test/Air/AirParser.test.js
```diff
@@ -1418,6 +1418,20 @@ describe('#AirParser', () => {
...
+    it('should parse booking with warnings and missing fare info and should show messages', () => {
+      const uParser = new Parser('universal:UniversalRecordImportRsp', 'v36_0', { });
+      const parseFunction = airParser.AIR_CREATE_RESERVATION_REQUEST;
+      const xml = fs.readFileSync(`${xmlFolder}/UniversalRecordWithWarnings.xml`).toString();
+      return uParser.parse(xml)
+        .then(json => parseFunction.call(uParser, json))
+        .then((result) => {
+          expect(result[0].fareQuotes.length).to.be.eq(0);
+          expect(result[0].messages.length).to.be.eq(2);
+          expect(result[0].messages[0]._).to.be.eq('Reservation is currently being updated by another process.');
```
shmuga: Is it going to be an array in case when only one message returned? Cause parser used under the hood could collapse things. I would probably add one more test for the case with one message

src/Services/Air/AirParser.js
```diff
@@ -755,14 +761,18 @@ function extractBookings(obj) {
...
-          const uapiSegmentRefs = pricingInfo['air:BookingInfo'].map(
+          const uapiSegmentRefs = (pricingInfo['air:BookingInfo'] || []).map(
             segment => segment.SegmentRef
           );
 
           const uapiPassengerRefs = pricingInfo[`common_${this.uapi_version}:BookingTravelerRef`];
 
           const fareInfo = pricingInfo['air:FareInfo'];
 
+          if (!fareInfo) {
```
shmuga: Is it fine that taxes, passengers won't be parsed when fare info is missing. It seem they are parsed from another structure

test/Air/AirParser.test.js
cursedcoder: it won't https://github.com/Travelport-Ukraine/uapi-json/blob/master/src/Request/default-config.js#L19

src/Services/Air/AirParser.js
cursedcoder: should be ok now

PR #424, opened 2019-07-24
Fix wrong stopovers once again

cursedcoder:
I just found this VERY STRANGE commit pushed to master straightforwardly https://github.com/Travelport-Ukraine/uapi-json/commit/c6f17a38aa367e08d7a1af21787c075a0942b571

Seems to be a wrong condition.

Added gds terminal FB mask to the test so you have to be sure about that.

### DIFF
--- a/src/Services/Air/AirParser.js
+++ b/src/Services/Air/AirParser.js
@@ -346,12 +346,34 @@ function getTicketFromEtr(etr, obj) {
     ? utils.firstInObj(airPricingInfo['air:FareInfo'])
     : null;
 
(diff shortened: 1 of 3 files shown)


src/Services/Air/AirParser.js
```diff
@@ -401,7 +401,7 @@ function getTicketFromEtr(etr, obj) {
               stopover: (
                 nextCoupon
                   ? nextCoupon.StopoverCode === 'true'
-                  : true
+                  : false
```
dchertousov: If there is no next Coupon, this is the last one, so stopOver should be true

PR #426, opened 2019-08-02
[NPM] Updates packages and fixes tests noise

shmuga:
### Scope 
Some of the packages were not updated last time due to breaking
changes. This PR updates them and fixes broken tests.

### Changes
- Adds .editorconfig
- Updates package.json and fixes sinon breaking changes
- Adds setupTests and stubs console.log to reduce noise in tests.
- Fixes tests in AirParser with uncaught errors


test/setupTests.test.js
```diff
@@ -0,0 +1,7 @@
+const sinon = require('sinon');
+
+before(() => {
```
cursedcoder: would be nice to have in-code explanation for the purpose, or you can use magic of naming and have `setupNoiseReduction.js` or similars ;)

shmuga: I've added a comment about this stub. Will leave filename to be used as single source of global test configurations.

PR #436, opened 2019-10-22
Prefix is missing in PNR

creduo:
in AirCreateReservationReq/BookingTraveler/BookingTravelerName, fixed misplaced condition


src/Services/Air/templates/AIR_CREATE_RESERVATION_REQUEST.handlebars.js
```diff
@@ -24,7 +24,7 @@ module.exports = `
             {{/if}}
             {{#each passengers}}
             <com:BookingTraveler Key="P_{{@index}}" Age="{{Age}}" DOB="{{DOB}}" Gender="{{gender}}" TravelerType="{{ageCategory}}">
-                <com:BookingTravelerName First="{{firstName}}" Last="{{lastName}}" {{#if Prefix}}Prefix="{{title}}"{{/if}}/>
+                <com:BookingTravelerName First="{{firstName}}" Last="{{lastName}}" {{#if Prefix}}Prefix="{{Prefix}}"{{/if}}/>
```
dchertousov: Let's change it to `title`.

Documentation should be updated as well, so users will be aware, they could use it

creduo: Changed to `title` as you commented, also updated Air.md.

PR #439, opened 2019-11-12
Issue-438 Segments formatted wrongly

JairoPanduro:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "1.2.3",
+  "version": "1.2.4",
```
cursedcoder: This should be `1.3.0` as you are making a breaking change.

cursedcoder:
Nice first-time contribution @JairoPanduro
Thank you!

PR #443, opened 2019-11-20
Add status codes to errors list, bump errors helpers

cursedcoder:


src/Services/Air/AirErrors.js
```diff
@@ -41,7 +41,7 @@ Object.assign(AirValidationError, createErrorsList({
 
 const GdsValidationError = createErrorClass(
   'GdsValidationError',
-  'Gds service validation error',
+  ['Gds service validation error', 200],
```
shmuga: Does it stand for statu code in here?

dchertousov: @shmuga +1

PR #447, opened 2019-12-05
TCP-2006 | Add status codes in terminal errors

cursedcoder:


src/Services/Air/Air.js
```diff
@@ -86,11 +86,21 @@ module.exports = (settings) => {
         });
     },
 
-    importPNR(options) {
-      return this.getPNR(options)
+    getPnr(options) {
```
InkFaust: `getPNR`, not `getPnr`

src/Services/Air/Air.js
```diff
@@ -350,24 +365,28 @@ module.exports = (settings) => {
...
-            .then(booking => service.cancelPNR(booking))
+            .then(() => this.getBooking(options))
+            .then(booking => service.cancelBooking(booking))
             .catch(
               err => Promise.reject(new AirRuntimeError.FailedToCancelPnr(options, err))
             );
         });
     },
 
+    cancelPNR(options) {
```
InkFaust: Deprecation warning?

PR #448, opened 2019-12-06
handle-travelport-server-error

hectorbenitez:
If travelport is having a `General air service Error.` it returns an error without err.data['universal:UniversalRecord']

`{ 
     faultcode: 'Server.Business',
     faultstring: 'General air service Error.',
     detail: { 'air:AvailabilityErrorInfo': [Object] } }
 }`


cursedcoder:
Seems that tests are failing.

It would be good to have test coverage for this as well.

PR #458, opened 2020-02-22
 Reference Data Request added and Some link errors corrected

jthoma:
[Reference data](https://support.travelport.com/webhelp/uapi/Content/Getting_Started/Design_Considerations/Reference_Data.htm) includes a number of categories of encode/decode data that contain mappings between human-readable, travel-industry names and codes recognized by Universal API. This data is also used internally within Universal API for validation. 

Checked the whole implementation to see if this is implemented, and found that it is not. Hence did the implmentation and tested using some of the data. Lint check of files created or edited by me was corrected, though there are some warnings on other files.

Also while editing the docs/Utils.md, found that the example link for "Currency convert" was missing the ".js"  and this is also updated.


docs/Utils.md
```diff
@@ -6,9 +6,10 @@ Some powerfull utils that can help you build best products.
 
 **UtilsService**
 * [.currencyConvert(params)](#currency)
+* [.referenceDataReq(params)](#refdata)
```
shmuga: The name could probably be simplified to `referenceData` as it's obvious that this will be a req :)

docs/Utils.md
```diff
@@ -18,4 +19,18 @@ Return list of currencies exchange rates.
...
+<a name="refdata"></a>
+## .referenceDataReq(params)
+Returns list of Reference Data 
+
+**Returns**: `Promise`
+**See**: [Reference Data](https://support.travelport.com/webhelp/uapi/Content/Getting_Started/Design_Considerations/Reference_Data.htm)
+
+| Param | Type | Description |
+| --- | --- | --- |
+| DataType | `String` | Types of Reference Data as specified from the support page. <a href="./ReferenceDataTypes.md">See List</a> |
```
shmuga: It's better to keep same naming convention for param names being camelCase instead of PascalCase

examples/Utils/referenceDataReq.js
```diff
@@ -0,0 +1,29 @@
+const uAPI = require('../../index');
+const config = require('../../test/testconfig');
+
+function uuidv4() {
```
shmuga: Probably https://www.npmjs.com/package/uuid in devDependencies won't harm

test/Utils/UtilsParser.test.js
```diff
@@ -62,4 +62,59 @@ describe('#utilsParser', () => {
...
   });
+  describe('referenceDataReq()', () => {
+    it('should parse sample data response for Country', () => {
+      const uParser = new Parser('util:ReferenceDataRetrieveRsp', 'v47_0', {});
+      const parseFunction = utilsParser.REFERENCE_DATATYPE;
+      const xml = fs.readFileSync(`${xmlFolder}/country-data.xml`).toString();
+      return uParser.parse(xml).then((json) => {
+        const result = parseFunction.call(uParser, json);
+        expect(result).to.be.an('array').and.to.have.lengthOf(3);
+        result.forEach(c => expect(c).to.be.an('object').and.to.have.all.keys(['Code', 'Name', 'Deprecated']));
```
shmuga: Nice one! Could we transform keys to camelCase as well? So resulting object would have next shape:

```
type Response = ReferenceData[];

type ReferenceData = {
  name: string,
  code: string,
  deprecated: boolean
}
```

src/reference-data-types.js
```diff
@@ -0,0 +1,43 @@
+module.exports = [
```
shmuga: Do all of these types return `{ name, code, deprecated }`?

docs/Utils.md
jthoma: Sure
This is done and pushed

docs/Utils.md
jthoma: Absolutely.. current commit is done with the changes

examples/Utils/referenceDataReq.js
jthoma: Yea.. did not notice that was there

src/reference-data-types.js
jthoma: Yes and no, there may be more properties.

test/Utils/UtilsParser.test.js
jthoma: Parser updated to dynamically rewrite the keys to camelCase

src/reference-data-types.js
jthoma: For VehicleSupplierType it is
`{
	"Code": "CI",
	"Description": "Choice International - Comfort Inns",
	"Deprecated": "false"
}`

src/reference-data-types.js
```diff
@@ -0,0 +1,43 @@
...
+  'BusinessType',
+  //  'Carrier', //
+  'City',
+  'CityAirport',
+  'Country',
+  'Currency',
+  'EmailType',
+  //  'Equipment', //
+  'FulfillmentType',
+  //  'GeoPoliticalArea', //
```
shmuga: What is the purpose of having commented types in here?

src/Services/Utils/UtilsParser.js
```diff
@@ -20,9 +20,29 @@ function currencyConvertParse(json) {
...
 }
 
+function dataTypeParse(json) {
+  json = json['util:ReferenceDataItem'];
+
+  if (Object.prototype.toString.call(json) !== '[object Array]') {
+    throw new UtilsParsingError(json);
+  }
+
+  const props = Object.getOwnPropertyNames(json[0]);
```
shmuga: This does look a bit unsafe. Image first element will have { code, deprecated } only and some of them as you mentioned previously (and [docs as well](https://support.travelport.com/webhelp/uapi/uAPI.htm#../Subsystems/Schemas/Content/Schemas/ReferenceDataRetrieveRsp.html%3FTocPath%3DSchema%7C_____186) ) might have `name` or `description`.

I would probably write it in a way where we handle all cases like

```javascript
const knownFieds = ['Code', 'Deprecated'];
const optionalFields = ['Name', 'Description'];

json = json.map((referenceItem) => {
  const item = {};
  knownFields.forEach(...);

  optionalFields.forEach(field => if (referenceItem[field]) { ... } )

  return item;
   
});
```

jthoma: Yes.. it seems more predicatble.. will do

src/reference-data-types.js
jthoma: Well those are supported only through referenceDataSearch. Think it would be better to move into another file ?

jthoma: Will keep those in my backup dir till I complete integration of referenceDataSearch

test/Utils/UtilsParser.test.js
```diff
@@ -62,4 +62,81 @@ describe('#utilsParser', () => {
...
+        const result = parseFunction.call(uParser, json);
+        expect(result).to.be.an('array').and.to.have.lengthOf(3);
+        result.forEach(c => expect(c).to.be.an('object').and.to.have.all.keys(['code', 'name', 'description', 'deprecated', 'additionalInfo']));
+      });
+    });
+
+    it('should throw parsing error', () => {
+      const parseFunction = utilsParser.REFERENCE_DATATYPE;
+      try {
+        parseFunction({});
```
cursedcoder: If the function doesn't throw, then the test will be green

src/Services/Utils/UtilsValidator.js
```diff
@@ -27,11 +28,31 @@ Validator.prototype.currencies = function () {
...
+Validator.prototype.datatype = function () {
+  if (Object.prototype.toString.call(this.params.dataType) !== '[object String]') {
+    throw new UtilsValidationError.DataTypeMissing(this.params);
+  }
+
+  if (this.params.dataType.length <= 0) {
+    throw new UtilsValidationError.DataTypeMissing(this.params);
+  }
+
+  if (referencedatatypes.indexOf(this.params.dataType) === -1) {
```
cursedcoder: let's use this fancy `if (!array.includes(item)` syntax, what do you think?

src/Services/Utils/UtilsValidator.js
```diff
@@ -1,4 +1,5 @@
 const { UtilsValidationError } = require('./UtilsErrors');
+const referencedatatypes = require('../../reference-data-types');
```
cursedcoder: nice validator! let's make it more readable by using snakeCase?

src/Services/Utils/UtilsParser.js
```diff
@@ -20,9 +21,50 @@ function currencyConvertParse(json) {
...
+
+  json = json.map((referenceItem) => {
+    const item = {};
+
+    if (Object.prototype.toString.call(referenceItem.Code) === '[object String]') {
+      item.code = referenceItem.Code;
+    }
+    if (Object.prototype.toString.call(referenceItem.Deprecated) === '[object String]') {
+      if (referenceItem.Deprecated === 'false') {
+        item.deprecated = false;
```
cursedcoder: simply `item.deprecated = referenceItem.Deprecated === 'false'`, 5 lines vs 1 , let's save some bytes

cursedcoder:
Many thanks by submitting, please check up a few comments.

test/Utils/UtilsParser.test.js
jthoma: Did not get what you meant. Actually it does throw since null object does not have any more properties, and that is checked in line 27 of src/Services/Utils/UtilsParser.js

src/Services/Utils/UtilsValidator.js
jthoma: Made it that way..

src/Services/Utils/UtilsValidator.js
jthoma: Did you mean camelCase ? I made that way

src/Services/Utils/UtilsParser.js
jthoma: Aah! yes this will save those...

test/Utils/UtilsParser.test.js
jthoma: ;) hmm.. made some changes.. should.fail added to the end after try..catch

Issue #461, opened 2020-03-03
AirService.cancelBooking always fails


jthoma:
When attempting to cancel a booking using the following syntax 
```
const params = {
  pnr: '5D1T2M',
  cancelTickets: true
};

AirService.cancelBooking(params).then(
  data => console.log(data),
  err => console.log(err)
);
```
I am getting an error as below

AirRuntimeError.UnableToRetrieveTickets: Unable to retrieve tickets list

PR #462, opened 2020-03-03
Closes issue AirService.cancelBooking always fails #461

jthoma:
Found the bug causing issue #461 and this PR will close the issue

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -350,7 +350,10 @@ module.exports = (settings) => {
 
       return this.getUniversalRecordByPNR(options)
         .then((ur) => {
-          const record = Array.isArray(ur) ? ur[0] : ur;
+          ur = Array.isArray(ur) ? ur[0] : ur;
+          const record = {
+            reservationLocatorCode: ur.uapi_reservation_locator
+          };
           return (ignoreTickets
             ? Promise.resolve([])
             : this.getTickets(record).then(checkTickets)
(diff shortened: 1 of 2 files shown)


jthoma:
@cursedcoder 
Please review, the PR

src/Services/Air/Air.js
```diff
@@ -350,7 +350,10 @@ module.exports = (settings) => {
 
       return this.getUniversalRecordByPNR(options)
         .then((ur) => {
-          const record = Array.isArray(ur) ? ur[0] : ur;
+          ur = Array.isArray(ur) ? ur[0] : ur;
```
dchertousov: Unneeded mutation

src/Services/Air/AirService.js
```diff
@@ -209,7 +209,7 @@ module.exports = function (settings) {
       debug,
       options
     ),
-    cancelPNR: uApiRequest(
+    cancelBooking: uApiRequest(
```
dchertousov: Nice rename

src/Services/Air/Air.js
jthoma: @dchertousov, could you explain your observation ?

I was just keen on getting this to work, since I am already deploying the wrapper on AWS Serverless using **sam** and **cloudformation** . After the change, the cancellation is working smoothly.

cursedcoder: I think @dchertousov means you can avoid mutating the parameter, instead declare a new variable.

PR #464, opened 2020-03-04
Replacing all ";" with "\t" when sending command

dchertousov:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "1.4.0",
+  "version": "1.3.3",
```
cursedcoder: Welp, it is not like that.
Please leave as it is.

PR #469, opened 2020-03-18
Implemented Schedule Change Accept / AckScheduleChangeReq/Rsp

creduo:
Due to recent global event, there are too many cancellations and schedule changes, this seems needed in some cases.

calling AckScheduleChangeReq will update seat statuses from TK to HK.


docs/Air.md
```diff
@@ -26,6 +26,7 @@ The Air workflow allows you to do what most travel agents did in the past and wh
 * [.getTickets(params)](#getTickets)
 * [.cancelTicket(params)](#cancelTicket)
 * [.cancelBooking(params)](#cancelBooking)
+* [.ackScheduleChange(params)](#ackScheduleChange)
```
dchertousov, line 29: Please rename to `acknowledge` or `accept` to make it more readable.
I needed to consult vocabulary for abbreviation meaning :)

src/Services/Air/AirParser.js
```diff
@@ -1327,6 +1332,7 @@ module.exports = {
   AIR_GET_TICKETS_ERROR_HANDLER: airGetTicketsErrorHandler,
   AIR_CANCEL_TICKET: airCancelTicket,
   AIR_CANCEL_PNR: airCancelPnr,
+  ACK_SCHEDULE_CHANGE: ackScheduleChange,
```
dchertousov, line 1335: should be still prepended by `AIR`

docs/Air.md
creduo: Accept sounds more natural, but the uAPI named ack, so renamed to acknowledgeScheduleChange

src/Services/Air/AirParser.js
creduo: Added AIR_ prefix

docs/Air.md
dchertousov: We don't always follow api naming routine, you know :)
In your another PR you took traveler as it's in uAPI, but we had passengers already, so you may keep accept. I find it more suitable too.

creduo: So, should we name it to acceptScheduleChange?

dchertousov: I would prefer it.

creduo: renamed it to acceptScheduleChange :)

PR #470, opened 2020-03-19
Codeshare information in LFS Result

creduo:
This pr adds codeshare informations to segments in LFS Result

eg.
> "isCodeshare": true,
"operatingCarrier": "FM",
"operatingCarrierName": "SHANGHAI AIRLINES CO LTD",
"operatingFlightNumber": "829",


src/Services/Air/AirFormat.js
```diff
@@ -45,6 +56,8 @@ function formatSegment(segment) {
     airline: segment.Carrier,
     flightNumber: segment.FlightNumber,
     uapi_segment_ref: segment.Key,
+    codeshare,
```
shmuga: I would name it `isCodeshare` to indicate that it's a boolean flag

creduo: changed to isCodeshare

src/Services/Air/AirFormat.js
```diff
@@ -86,6 +86,17 @@ function getBaggageInfo(info) {
 }
 
 function formatSegment(segment) {
+  const isCodeshare = (Object.prototype.toString.call(segment['air:CodeshareInfo']) === '[object Object]');
+  let codeshareInfo = {};
```
dchertousov, line 90: Would prefer `const`

```javascript
const codeshareInfo = isCodeshare
  ? {
    operatingCarrier: segment['air:CodeshareInfo'].OperatingCarrier,
    operatingCarrierName: segment['air:CodeshareInfo']._,
    operatingFlightNumber: segment['air:CodeshareInfo'].OperatingFlightNumber,
  }
  : {};
```

dchertousov: With let it's always easy to mutate when it's not needed to.

src/Services/Air/AirFormat.js
```diff
@@ -86,6 +86,17 @@ function getBaggageInfo(info) {
...
+
+  if (isCodeshare) {
+    codeshareInfo = {
+      operatingCarrier: segment['air:CodeshareInfo'].OperatingCarrier,
+      operatingCarrierName: segment['air:CodeshareInfo']._,
```
dchertousov, line 95: Will it be the full name of carrier or what?

src/Services/Air/AirFormat.js
```diff
@@ -95,6 +106,8 @@ function formatSegment(segment) {
     airline: segment.Carrier,
     flightNumber: segment.FlightNumber,
     uapi_segment_ref: segment.Key,
+    isCodeshare,
+    ...codeshareInfo
```
dchertousov, line 110: I would either to keep codeshareInfo as structure here or to prepend codeshare keys with `codeshare`.
First option sounds more preferable to me.

src/Services/Air/AirFormat.js
creduo: This will be full name of carrier -ish.

src/Services/Air/AirFormat.js
creduo: Used Object.assign to add codeshare informations when isCodeshare is true

PR #471, opened 2020-03-23
Implemented PNR divide

creduo:
This is an implementation of UniversalService/ProviderReservationDivideReq.

### DIFF
--- a/docs/Air.md
+++ b/docs/Air.md
@@ -434,3 +434,21 @@ Gets pnr information and tickets list from [`importPNR`](#importPNR) and then do
 | ignoreTickets | `Boolean` | Defines if tickets should be ignored. The default value is `false` |
 
 **See: <a href="../examples/Air/cancelBooking.js">cancelBooking example</a>**
+
+## .providerReservationDivide(params)
+<a name="providerReservationDivide"></a>
+
+Sends a pnr divide request and returns parent and child reservation information with divided passenger list.
+
+**Returns**: `Promise` 
+
+**See**: [Dividing Universal Records](https://support.travelport.com/webhelp/uapi/uapi.htm#Booking/UniversalRecord/Dividing_Universal_Records.htm%3FTocPath%3DUniversal%2520Records%2520and%2520Bookings%7C_____5)
+and <a href="../examples/Air/divide.js">providerReservationDivide example</a>**
+
+| Param | Type | Description |
+| --- | --- | --- |
+| uapi_ur_locator | `Object` | Universal Record Locator |
+| provider | `String` | GDS provider code eg. '1G' |
+| pnr | `String` | PNR |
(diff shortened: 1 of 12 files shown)


creduo:
This PR may contain breaking changes as BookingTravelerName is added in the noCollapseList. Please review carefully regarding this change.

src/Services/Air/AirParser.js
```diff
@@ -1084,6 +1084,29 @@ function universalRecordRetrieveRequest(data) {
   return response;
 }
 
+function formatProviderReservationInfo(data) {
+  return {
+    traveler: data['common_v47_0:BookingTravelerName'].map((travelerDetails) => {
```
dchertousov, line 1089: Would prefer to call it passengers, as we never used `traveler` before and it is array.

test/Air/Air.test.js
```diff
@@ -1740,6 +1740,26 @@ describe('#AirService', () => {
...
+      const createAirService = proxyquire('../../src/Services/Air/Air', {
+        './AirService': service,
+      });
+      createAirService({ auth }).providerReservationDivide({
+        uapi_ur_locator: 'AAABBB',
+        version: '1',
+        provider: '1G',
+        pnr: '456ABC',
+        passengers: [
+          { uapi_ref_key: 'XX/ADFH2345AGJDS245IGJSDifgjoai22343s==' },
```
dchertousov, line 1756: It's `uapi_passenger_ref` in example and templates. How could this test work?

creduo: renamed it to uapi_passenger_ref last minute, this must have been left out

docs/Air.md
```diff
@@ -437,6 +437,23 @@ Gets pnr information and tickets list from [`importPNR`](#importPNR) and then do
...
+Sends a pnr divide request and returns parent and child reservation information with divided passenger list.
+
+**Returns**: `Promise` 
+
+**See**: [Dividing Universal Records](https://support.travelport.com/webhelp/uapi/uapi.htm#Booking/UniversalRecord/Dividing_Universal_Records.htm%3FTocPath%3DUniversal%2520Records%2520and%2520Bookings%7C_____5)
+and <a href="../examples/Air/divide.js">providerReservationDivide example</a>**
+
+| Param | Type | Description |
+| --- | --- | --- |
+| uapi_ur_locator | `Object` | Universal Record Locator |
```
dchertousov, line 452: It's not an object.
Also, could we allow to skip passing this locator as w we might get it internally?

dchertousov: And could you please add a description of the response?

dchertousov (changes requested):
Thank you for that very helpful method!

Just few things are to be enhanced, and we are ready to roll with it

PR #472, opened 2020-03-24
Fixed AirPricingModifier in templates 

creduo:
Fixed business and platingCarrier was wrongly placed with plating carrier, and added few more attributes.


src/Services/Air/templates/AIR_PRICE_REQ.handlebars.js
```diff
@@ -37,17 +37,31 @@ module.exports = `
                 </air:AirSegment>
                 {{/segments}}
             </air:AirItinerary>
-            {{#if platingCarrier}}
-              <air:AirPricingModifiers PlatingCarrier="{{platingCarrier}}"/>
-            {{/if}}
+            
+            <air:AirPricingModifiers
+                InventoryRequestType="DirectAccess"
```
dchertousov, line 42: Should it always be DirectAccess?

creduo: It should bes to avoid unreliable inventory. Although this does not work with 1G, this only for 1P.

src/Services/Air/templates/AIR_PRICE_REQ.handlebars.js
```diff
@@ -37,17 +37,31 @@ module.exports = `
...
+                FaresIndicator="{{pricing.faresIndicator}}"
+                {{else}}
+                FaresIndicator="PublicAndPrivateFares"
+                {{/if}}
             {{#if business}}
-            <air:AirPricingModifiers InventoryRequestType="DirectAccess">
+            >
                 <air:PermittedCabins>
-                    <com:CabinClass Type="Business" xmlns:com="http://www.travelport.com/schema/common_v47_0" />
+                    <com:CabinClass Type="Business" />
```
dchertousov, line 60: Could you please take a look at #481 and apply permitted cabins based on `permittedCabins` option?
If not, then we could create another feature request for it.

creduo: Added cabin options similar to LFS and Availability, so we can specify Economy or First Class, with legacy "business". But to change this adequately, maybe more discussion would be required.

PR #476, opened 2020-04-14
Parse records with missing itinerary

InkFaust:


src/Services/Air/AirParser.js
```diff
@@ -778,6 +800,34 @@ function extractBookings(obj) {
       )
     : {};
 
+  if (!hasAirReservation) {
+    return Object.keys(reservationInfo).map((key) => {
+      const providerInfo = reservationInfo[key];
+      const passengers = Object.keys(record['common_v47_0:BookingTraveler']).map((travelerKey) => {
```
dchertousov, line 806: Please fix reference to versions. Should not be hardcoded

src/Services/Air/AirParser.js
```diff
@@ -735,6 +735,25 @@ function airCancelPnr(obj) {
   throw new AirParsingError.CancelResponseNotFound();
 }
 
+function buildPassenger(name, traveler) {
```
dchertousov, line 738: Move to AirFormat

PR #477, opened 2020-04-23
TCP 2146

InkFaust:


docs/Air.md
```diff
@@ -434,3 +435,19 @@ Gets pnr information and tickets list from [`importPNR`](#importPNR) and then do
...
+
+Add segments to an existing reservation record.
+
+**Returns**: `Promise` which is resolved with response message.
+
+| Param | Type | Description |
+| --- | --- | --- |
+| pnr | `String` | PNR |
+| version | `Number` | Current uAPI record version |
+| universalRecordLocatorCode | `String` | uAPI universal locator code |
```
dchertousov, line 450: Are `universalRecordLocatorCode` and `reservationLocatorCode` retrievable by PNR?
If yes I would like to make them optional. If they are not provided, would be fin to extract them on our side.

docs/Air.md
```diff
@@ -434,3 +435,19 @@ Gets pnr information and tickets list from [`importPNR`](#importPNR) and then do
...
+| reservationLocatorCode | `String` | uAPI reservation locator code |
+| bookingTravelerRef | `String` | uAPI booking traveler reference key |
```
dchertousov, line 452: Have you tried adding segments to the PNR with more than one passenger?
I believe, segments could not be added per passenger.
Please check and take not the following actions:
- If multiple references should be provided, change type to array
- If one reference could be provided, please report

Either way if we could retrieve it by PNR as well as data in previous comment, would be fine to keep this params optional

InkFaust: Figured out it's an optional parameter and can be safely skipped, I'll create a new booking and check it to confirm.

examples/Air/addSegments.js
```diff
@@ -0,0 +1,36 @@
...
+    production: true,
+  }
+);
+
+
+const params = {
+  pnr: 'PNR000',
+  version: 5, // Optional
+  universalRecordLocatorCode: 'URLC00', // Optional
+  reservationLocatorCode: 'RLC00', // Optionals
```
dchertousov, line 17: => Optional

docs/Air.md
```diff
@@ -434,3 +435,18 @@ Gets pnr information and tickets list from [`importPNR`](#importPNR) and then do
 | ignoreTickets | `Boolean` | Defines if tickets should be ignored. The default value is `false` |
 
 **See: <a href="../examples/Air/cancelBooking.js">cancelBooking example</a>**
+
+## .addSegments(params)
```
dchertousov, line 439: Please, add a statement, that version should be passed if user wants to make sure, that he had the last version overview before change. Could be like:
"It is recommended to pass `version` param in order to ensure that last version of the PNR was reviewed before modification".

src/Services/Air/Air.js
```diff
@@ -45,6 +45,31 @@ module.exports = (settings) => {
...
+              version: options.version || booking.version,
+              reservationLocatorCode: (
+                options.reservationLocatorCode || booking.uapi_reservation_locator
+              ),
+              universalRecordLocatorCode: (
+                options.universalRecordLocatorCode || booking.uapi_ur_locator
+              ),
+            };
+
+            return service.addSegments(Object.assign(options, missedOptions));
```
dchertousov, line 66: `Object.assign` is not necessary anymore, while you combine and object from single values above.
Also, `Object.assign` mutates first param.

PR #490, opened 2020-09-18
Fix gender in passenger format

kim00425:
Extract gender from infant last name

### DIFF
--- a/src/Services/Air/AirFormat.js
+++ b/src/Services/Air/AirFormat.js
@@ -515,12 +515,14 @@ function setIndexesForSegments(
 }
 
 function buildPassenger(name, traveler) {
-  if (traveler.TravelerType === 'INF') {
-    const infantGender = name.First.match(/MSTR|MISS$/gi);
+  if (traveler.TravelerType && traveler.TravelerType === 'INF') {
+    if (name.First && String.isString(name.First)) {
+      const infantGender = name.First.match(/MSTR|MISS$/gi);
 
-    if (infantGender && (!traveler.Gender)) {
-      name.First = name.First.replace(/MSTR|MISS$/gi, '');
-      traveler.Gender = infantGender === 'MSTR' ? 'M' : 'F';
+      if (infantGender && (!traveler.Gender)) {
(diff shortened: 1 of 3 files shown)


kim00425:
> Extract gender from infant last name

dchertousov (changes requested):
Thanks for your PR!

In case of non-INF Gender could be missing (as still could be parsed from name), but for the INF passengers you propose to always include gender based on First name.

IMO, we should to stick for the similar behaviour in all cases.

Also we try not to mutate function params, so I would like to see the code slightly modified:

- if gender field exists, always take the gender from it
- if gender field is absent, take the gender from the name
- if name does not include strings like MR, RS, MSTR MISS etc, set gender to null
- use separate variable when detecting gender, using rules above instead of modifying function params

Please also make sure, you have included all possible variants of MR/MRS strings and your reg exp is valid.

Current regex has some flaws:

```
(/MSTR|MISS$/gi).test('MSTRME/NAME')
````
this one will return true, while it should not. This would return false correctly:

```
(/(?:MSTR|MISS)$/gi).test('MSTRME/NAME')
```

kim00425:
> Thanks for your PR!
> 
> In case of non-INF Gender could be missing (as still could be parsed from name), but for the INF passengers you propose to always include gender based on First name.
> 
> IMO, we should to stick for the similar behaviour in all cases.
> 
> Also we try not to mutate function params, so I would like to see the code slightly modified:
> 
> * if gender field exists, always take the gender from it
> * if gender field is absent, take the gender from the name
> * if name does not include strings like MR, RS, MSTR MISS etc, set gender to null
> * use separate variable when detecting gender, using rules above instead of modifying function params
> 
> Please also make sure, you have included all possible variants of MR/MRS strings and your reg exp is valid.
> 
> Current regex has some flaws:
> 
> ```
> (/MSTR|MISS$/gi).test('MSTRME/NAME')
> ```
> 
> this one will return true, while it should not. This would return false correctly:
> 
> ```
> (/(?:MSTR|MISS)$/gi).test('MSTRME/NAME')
> ```

i did PR again check please

PR #491, opened 2020-09-28
Working on NextResultReference in air availability

dchertousov:
- NextResultReference fixed
- Dependencies updated

### DIFF
--- a/src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
+++ b/src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
@@ -11,9 +11,7 @@ module.exports = `
                 <com:OverridePCC ProviderCode="{{provider}}" PseudoCityCode="{{emulatePcc}}"/>
             {{/if}}
             {{#nextResultReference}}
-                <com:NextResultReference ProviderCode="{{provider}}" >
-                    {{nextResultReference}}
-                </com:NextResultReference>
+                <com:NextResultReference ProviderCode="{{../provider}}">{{{.}}}</com:NextResultReference>
             {{/nextResultReference}}
             {{#legs}}
             <air:SearchAirLeg>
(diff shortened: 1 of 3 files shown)


src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
```diff
@@ -11,9 +11,7 @@ module.exports = `
                 <com:OverridePCC ProviderCode="{{provider}}" PseudoCityCode="{{emulatePcc}}"/>
```
InkFaust, line 11: Shouldn't it also have `../provider` updated?

src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
```diff
@@ -11,9 +11,7 @@ module.exports = `
...
             {{/if}}
             {{#nextResultReference}}
-                <com:NextResultReference ProviderCode="{{provider}}" >
-                    {{nextResultReference}}
-                </com:NextResultReference>
+                <com:NextResultReference ProviderCode="{{../provider}}">{{{.}}}</com:NextResultReference>
```
InkFaust, line 14: It's not self-explanatory what the result of `{{{.}}}` will be, could you elaborate on that, please?

src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
dchertousov: No, as this part uses `if` statement instead of going into context. Tried the same with `nextResultReference`, but it didn't work

src/Services/Air/templates/AIR_AVAILABILTIY_REQUEST.handlebars.js
dchertousov: skip html escaping, as `nextResultReference` is a base64 string which might contain '=' symbol

InkFaust (approved):

PR #498, opened 2020-11-20
TC-1313 | Omit TicketInfoIncomplete error with a flag

InkFaust:


src/Request/uapi-request.js
```diff
@@ -181,7 +181,10 @@ module.exports = function uapiRequest(
       .then(sendRequest)
       .then(parseResponse)
       .then(validateSOAP)
-      .then(parseFunction.bind(uParser)) // TODO: merge Hotels
+      .then((res) => {
+        const parse = parseFunction.bind(uParser); // TODO: merge Hotels
```
dchertousov, line 185:
```javascript
(res) => parseFunction.call(uParser, res, params);
```

dchertousov (changes requested):
Please update node-error-helpers dependency and up versions.

dchertousov (approved):

PR #537, opened 2021-11-26
TC-2882: getUniversalRecordByPNR is using recordtRetrieve request now…

kommandant-topp:
…; importUniversalRecordByPNR created to use recordImport in case of need

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -125,6 +125,10 @@ module.exports = (settings) => {
       return service.getUniversalRecord(options);
     },
 
+    importUniversalRecordByPNR(options) {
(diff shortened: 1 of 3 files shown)


dchertousov:
@kommandant-topp Shouldn't we check for some error, while getting a record by PNR if it hasn't been imported yet and run import in that case?

kommandant-topp:
> @kommandant-topp Shouldn't we check for some error, while getting a record by PNR if it hasn't been imported yet and run import in that case?

I checked booking freshly created with smartpoint, and no error was returned. As Martin mentioned previously: "we usually use UniversalRecordRetrieve which works also in situation when the Universal record doesnt exist"

PR #538, opened 2021-11-29
TC-2860: -duplicate ticket found- error correct handling to get ticke…

kommandant-topp:
…t data

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -8,11 +8,6 @@ const createTerminalService = require('../Terminal/Terminal');
 const { AirRuntimeError } = require('./AirErrors');
 const validateServiceSettings = require('../../utils/validate-service-settings');
 
-const RETRYABLE_GET_TICKET_ERRORS = [
-  AirRuntimeError.TicketInfoIncomplete,
-  AirRuntimeError.DuplicateTicketFound,
-];
-
 module.exports = (settings) => {
   const service = airService(validateServiceSettings(settings));
   const log = (settings.options && settings.options.logFunction) || console.log;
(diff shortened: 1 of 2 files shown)


src/Services/Air/Air.js
```diff
@@ -259,6 +259,33 @@ module.exports = (settings) => {
...
 
+    async getTicketFromTicketsList(pnr, ticketNumber) {
+      const tickets = await this.getTickets({ pnr });
+      return tickets.find(t => t.ticketNumber === ticketNumber);
+    },
+
+    async getTicketDuplicateTicketFoundCase({
+      pnr,
+      ticketNumber,
+      allowNoProviderLocatorCodeRetrieval
```
dchertousov, line 270: This flag is not needed as you always provide `uapi_ur_locator` here

src/Services/Air/Air.js
```diff
@@ -259,6 +259,33 @@ module.exports = (settings) => {
...
+    }) {
+      const { splitBookings } = await this.getBooking({ pnr });
+      if (!splitBookings) {
+        return this.getTicketFromTicketsList(pnr, ticketNumber);
+      }
+
+      /* eslint-disable camelcase */
```
dchertousov, line 277: This way you disable it all the way down the file.

```
const { uapi_ur_locator: urLocator } = ...

return service.getTicket({
  uapi_ur_locator: urLocator,
  ...
});
```

src/Services/Air/Air.js
```diff
@@ -259,6 +259,33 @@ module.exports = (settings) => {
...
+      return tickets.find(t => t.ticketNumber === ticketNumber);
+    },
+
+    async getTicketDuplicateTicketFoundCase({
+      pnr,
+      ticketNumber,
+      allowNoProviderLocatorCodeRetrieval
+    }) {
+      const { splitBookings } = await this.getBooking({ pnr });
+      if (!splitBookings) {
```
dchertousov, line 273: As we discussed, this error should be thrown only if case of split booking. Instead of applying another workaround handler I'd like the code to throw an error when DuplicateTicket error is there, but no split bookings found.

I think it's better to return null here, so original error will be thrown

src/Services/Air/Air.js
```diff
@@ -269,8 +296,9 @@ module.exports = (settings) => {
         }
 
         const pnr = await this.getPNRByTicketNumber({ ticketNumber });
-        const tickets = await this.getTickets({ pnr });
-        const ticket = tickets.find(t => t.ticketNumber === ticketNumber);
+        const ticket = err instanceof AirRuntimeError.DuplicateTicketFound
```
dchertousov, line 299: Hard to expand this kind of handling to the third error.

```
const { helpers: { getFullName: getFullErrorName } } = require('../../node-errors-helpers');

retryableTicketErrorHandlers = {
  'AirRuntimeError.TicketInfoIncomplete': async (ticketNumber) => {
    const pnr = await this.getPNRByTicketNumber({ ticketNumber });
    const tickets = await this.getTickets({ pnr });
    return tickets.find(t => t.ticketNumber === ticketNumber);
  },
  'AirRuntimeError.DuplicateTicketFound': async (ticketNumber) => {
    const pnr = await this.getPNRByTicketNumber({ ticketNumber });
    const { splitBookings } = await this.getBooking({ pnr });
    if (!splitBookings) {
      return null;
    }
    const [splitBookingPNR] = splitBookings;
    const { uapi_ur_locator: urLocator } = await this.getBooking({ pnr: splitBookingPNR });
    return service.getTicket({
      ticketNumber,
      uapi_ur_locator: urLocator,
      pnr: splitBookingPNR,
    });
  },
}

async getTicket(options) {
  const { ticketNumber, allowNoProviderLocatorCodeRetrieval = false } = options;
  try {
    return await service.getTicket({ ticketNumber, allowNoProviderLocatorCodeRetrieval });
  } catch (err) {
    const errorName = getFullErrorName(err);
    const retryableErrorHandler = this.retryableTicketErrorHandlers[errorName];
    const ticket = await(retryableErrorHandler && retryableErrorHandler(ticketNumber));
    if (!ticket) {
      throw err;
    }
    return ticket
  }
}
```

Please use the approach similar to the described above, where you can register a handler for each error type

dchertousov (changes requested):
Please find my comments.
Also please bump a version to 1.10.2

PR #545, opened 2021-12-15
Tc 2903

kommandant-topp:


src/Services/Air/Air.js
```diff
@@ -266,19 +266,25 @@ module.exports = (settings) => {
       },
       'AirRuntimeError.DuplicateTicketFound': async function (ticketNumber) {
         const pnr = await this.getPNRByTicketNumber({ ticketNumber });
-        const { splitBookings } = await this.getBooking({ pnr });
-        if (!splitBookings) {
+
+        try {
           return this.getTicketFromTicketsList(pnr, ticketNumber);
```
dchertousov, line 271: Please add await here as otherwise just a promise will be returned, and the error will never be catch.

dchertousov: Please also add a test to catch and rethrow the error here.

dchertousov (changes requested):
See my comments

dchertousov (approved):

PR #547, opened 2021-12-24
TC-2912: update for handling several split bookings in case of Duplic…

kommandant-topp:
…ateTicketFound error

### DIFF
--- a/src/Services/Air/Air.js
+++ b/src/Services/Air/Air.js
@@ -267,22 +267,46 @@ module.exports = (settings) => {
       'AirRuntimeError.DuplicateTicketFound': async function (ticketNumber) {
         const pnr = await this.getPNRByTicketNumber({ ticketNumber });
         try {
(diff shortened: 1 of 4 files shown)


src/Services/Air/Air.js
```diff
@@ -267,22 +267,46 @@ module.exports = (settings) => {
...
         } catch (e) {
           const { splitBookings } = await this.getBooking({ pnr });
-
           if (!splitBookings) {
             throw e;
           }
 
-          const [splitBookingPNR] = splitBookings;
-          const { uapi_ur_locator: urLocator } = await this.getBooking({ pnr: splitBookingPNR });
+          return splitBookings.reduce(async (acc, splitPnr) => {
```
dchertousov, line 283: I'd like more to use map, return `null` if ticket does not match and then filter results. This way we could get results faster.

dchertousov (changes requested):
See my comments.

dchertousov (approved):

PR #570, opened 2022-03-24
TC-3010: md update to redo command in case of received session timeout

kommandant-topp:


src/Services/Terminal/Terminal.js
```diff
@@ -67,6 +73,57 @@ process.on('exit', () => {
...
+
+  return false;
+};
+
+const commandRetry = async ({
+  service,
+  command,
+  sessionToken,
+  times = DEFAULT_RETRY_TIMES,
+  count = 0
```
dchertousov, line 104: Unnecessary use of `count`. You may use times, gradually decreasing and throw an error when the `times = 0`

src/Services/Terminal/Terminal.js
```diff
@@ -67,6 +73,57 @@ process.on('exit', () => {
...
+}) => {
+  try {
+    const result = await service.executeCommand({ command, sessionToken });
+    return result;
+  } catch (e) {
+    const res = isErrorRetriable(e);
```
dchertousov, line 110:
```
if (!isErrorRetriable(e) || times <=0) {
  throw e;
}

return commandRetry({
  service,
  command,
  sessionToken,
  times: times - 1,
});
```

dchertousov (changes requested):
@kommandant-topp Please see my comments

src/Services/Terminal/Terminal.js
```diff
@@ -67,6 +73,57 @@ process.on('exit', () => {
...
+
+const commandRetry = async ({
+  service,
+  command,
+  sessionToken,
+  times = DEFAULT_RETRY_TIMES,
+  count = 0
+}) => {
+  try {
+    const result = await service.executeCommand({ command, sessionToken });
```
dchertousov, line 107: You may also use

```javascript
return await service.executeCommand({ command, sessionToken });
```

PR #573, opened 2022-04-04
air-reservation-req updated and added dobString if passenger is a child

ukeh:


dchertousov:
Thank you @ukeh!

Your code will be released in version 1.13.0

PR #581, opened 2022-07-14
TC-3037: retrieve ticketDesignator from AirPricingInfo

kommandant-topp:


src/utils/form-fare-basis-code.js
```diff
@@ -0,0 +1,21 @@
+module.exports = (airPricingInfo, coupon) => {
+  const couponTicketDesignator = coupon['air:TicketDesignator']
+    && coupon['air:TicketDesignator'].Value;
+
+  if (!airPricingInfo && couponTicketDesignator) {
```
dchertousov, line 5: IMO, nested if will be much easier to read

src/utils/form-fare-basis-code.js
```diff
@@ -0,0 +1,21 @@
...
+    return `${coupon.FareBasis}/${couponTicketDesignator}`;
+  } if (!airPricingInfo) {
+    return coupon.FareBasis;
+  }
+
+  const [fareInfoData] = Object.values(airPricingInfo['air:FareInfo'])
```
dchertousov, line 11: Is designator from `airPricingInfo` more prioritized?

src/utils/form-fare-basis-code.js
dchertousov: Or even ternary inside if s in the code below

dchertousov (commented):
Please see my comments.
Thanks

src/utils/form-fare-basis-code.js
kommandant-topp: correct, according to Martin's response: `I would say that you have to get AirPricingInfo and there parse FareInfo for each segment to get AirSegment/FareBasis/TicketDesignator. And AirSegments can be paired  with flight coupons` and test results

dchertousov (approved):

dchertousov (approved):

PR #586, opened 2022-07-26
TC-3242 / Better error handling

dchertousov:


kommandant-topp:
Could you create an error object { code: name } for the various error handler switches?

kommandant-topp (approved):

PR #598, opened 2022-08-30
Tc 3270

kommandant-topp:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "uapi-json",
-  "version": "1.14.0",
```
dchertousov, line 3: Please change minor instead of patch as this change is not backwards compatible and all users will start to receive other information for the same calls

dchertousov (changes requested):

dchertousov (approved):

PR #604, opened 2022-11-29
TC-3353 / getTickets improvements

dchertousov:
- Calling `getBooking` instead of `getUniversalRecordByPNR` for code reuse
- Now accepting `reservationLocator` param, so no need to call `getBooking` all the time
- UR check moved to `getBookingFromUr`
- Dependencies updated based on audit

### DIFF
--- a/src/utils/get-booking-from-ur.js
+++ b/src/utils/get-booking-from-ur.js
@@ -1,4 +1,8 @@
 module.exports = function getBookingFromUr(ur, pnr) {
+  if (!Array.isArray(ur)) {
+    return ur;
+  }
+
   return ur.find(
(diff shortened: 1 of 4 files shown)


src/utils/get-booking-from-ur.js
```diff
@@ -1,4 +1,8 @@
 module.exports = function getBookingFromUr(ur, pnr) {
+  if (!Array.isArray(ur)) {
```
kommandant-topp, line 2: Which case can we have it as not array?

kommandant-topp: never mind, found it

kommandant-topp (approved):

PR #614, opened 2023-01-09
TC-3363: correct air/service segments' indexes

kommandant-topp:


src/Services/Air/AirFormat.js
```diff
@@ -459,33 +467,20 @@ function setIndexesForSegments(
   segmentsObject = null,
   serviceSegmentsObject = null
 ) {
-  const segments = segmentsObject
```
dchertousov, line 462:
```

function setSegmentsIndexes(segments) {
  return segments.map((segment) => {
    const { ProviderSegmentOrder: index = null } = segment;
    if (index === null) {
      throw new Error('DDD');
    }
    return { ...segment, index };
  });
}

const segments = segments !== null
  ? setSegmentsIndexes(segments)
  : null;
const serviceSegments = segments !== null
  ? setSegmentsIndexes(serviceSegments)
  : null;

return { segments, serviceSegments };

File: src/Services/Air/AirFormat.js [hunk 2]
```diff
@@ -446,6 +438,24 @@ function formatLowFaresSearch(searchRequest, searchResult) {
   return fares;
 }
 
+const getSegmentsData = segmentsObject => (segmentsObject
+  ? Object.keys(segmentsObject).map(k => segmentsObject[k])
```
dchertousov, line 442: `Object.values` ?

dchertousov (changes requested):
Please see comment to the code (a really small one).
Please also update the version and make sure to include `npm audit` recommendations if there are any

package.json
```diff
@@ -32,7 +32,7 @@
     "Artem Pylypchuk <articicejuice@gmail.com>",
     "Yevhenii Huselietov <d46k16@gmail.com>",
     "Mark Omarov <dev.mark.omarov@gmail.com>",
-    "Oleksii Duvanov <kommandant@web-hata.com>"
+    "Oleksii Duvanov <colden.aid@gmail.com>"
```
dchertousov, line 35: 👍

dchertousov (approved):

PR #630, opened 2023-05-05
TC-3462: separate financial and comon unexpected Travelport errors

kommandant-topp:


dchertousov (changes requested):
Please see my comments

PR #649, opened 2024-02-27
Tc 3713

kommandant-topp:


dchertousov:
@kommandant-topp Could you please refactor the code, so the nesting will not be 4 levels deep?

